PR #12, opened 2017-07-06
Add /package path API methods

sudarsanm94:


presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoCommand.java
```diff
@@ -0,0 +1,20 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.teradata.prestomanager.agent;
+
+public interface PrestoCommand
+{
+    void runCommand()
+            throws Exception;
```
rschlussel: throw PrestoManagerException.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
```diff
@@ -13,66 +13,74 @@
...
 public final class PrestoInstaller
+        implements PrestoCommand
 {
-    private PrestoInstaller() {}
+    private static final int CONNECTION_TIMEOUT = 1000000;
+    private static final int READ_TIMEOUT = 10000;
 
-    public static void installPresto(String pathToRpm, boolean disableDependencyChecking)
-            throws PrestoManagerException
+    public final PackageType packageType;
```
rschlussel: All fields should be private

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUninstaller.java
```diff
@@ -0,0 +1,71 @@
...
+import com.teradata.prestomanager.agent.api.PackageAPI.PackageType;
+
+import static com.teradata.prestomanager.agent.CommandExecutor.executeCommand;
+import static java.lang.String.format;
+import static java.util.Objects.requireNonNull;
+
+public class PrestoUninstaller
+        implements PrestoCommand
+{
+    public final PackageType packageType;
```
rschlussel: make these private

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
```diff
@@ -0,0 +1,86 @@
...
+import static java.util.Objects.requireNonNull;
+import static org.apache.commons.io.FileUtils.copyURLToFile;
+
+public class PrestoUpgrader
+        implements PrestoCommand
+{
+    private static final int CONNECTION_TIMEOUT = 1000000;
+    private static final int READ_TIMEOUT = 10000;
+
+    public final PackageType packageType;
```
rschlussel: private

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -0,0 +1,145 @@
...
+    public class PrestoCommandAsynchronous
+            implements Runnable
+    {
+        PrestoCommand command;
+
+        public PrestoCommandAsynchronous(PrestoCommand command)
+        {
+            this.command = command;
+        }
+
```
rschlussel: annotate with @Overrides

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -0,0 +1,145 @@
...
+        new Thread(new PrestoCommandAsynchronous(new PrestoUninstaller(installedPackageType, disableDependencyChecking, ignoreNotInstalled))).start();
+        return Response.status(ACCEPTED).build();
+    }
+
+    public enum PackageType
+    {
+        RPM, TARBALL;
+    }
+
+    public class PrestoCommandAsynchronous
```
rschlussel: Call this PrestoCommandRunnable (or PrestoAsynchronousCommand... the main thing is that you end up with a noun).

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -0,0 +1,145 @@
...
+
+    @PUT
+    @Consumes({"text/plain"})
+    @Produces({"text/plain"})
+    @ApiOperation(value = "Install presto using rpm or tarball")
+    @ApiResponses(value = {
+            @ApiResponse(code = 202, message = "Acknowledged request"),
+            @ApiResponse(code = 400, message = "Bad package format or Invalid url")
+    })
+    public synchronized Response install(@NotNull String locationToFetchPackage,
```
rschlussel: you don't need an `@notnull` annotation since the default assumption is that you shouldn't be passing nulls

jirassimok:
I don't believe I ever finished my review of this PR.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -0,0 +1,122 @@
...
+import javax.ws.rs.POST;
+import javax.ws.rs.PUT;
+import javax.ws.rs.Path;
+import javax.ws.rs.QueryParam;
+import javax.ws.rs.core.Response;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+
+import static com.google.common.io.Files.getFileExtension;
```
jirassimok: @rschlussel That makes sense.
What about methods with descriptive but potentially misleading names, where qualifying the name does not help?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoCommand.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.teradata.prestomanager.agent;
+
+public interface PrestoCommand
```
jirassimok: Why not just extend `Runnable`? (if the answer is "exceptions", see my comment on that below)

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
```diff
@@ -13,66 +13,74 @@
...
     {
-        String checkRpm = format("rpm -Kv --nosignature %s", pathToRpm);
+        String checkRpm = format("rpm -K --nosignature %s", pathToRpm);
         if (executeCommand(checkRpm) != 0) {
             throw new PrestoManagerException("Corrupted RPM");
         }
         String nodeps = "";
         if (disableDependencyChecking) {
             nodeps = "--nodeps";
         }
```
jirassimok: The previous four lines can be simplified as `String nodeps = disableDependencyChecking ? "--nodeps" : ""`.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -0,0 +1,146 @@
...
+    {
+        try {
+            installedPackageType = PackageType.valueOf(getFileExtension(locationToFetchPackage).toUpperCase());
+        }
+        catch (IllegalArgumentException e) {
+            return Response.status(BAD_REQUEST).entity(text("Invalid package type for presto-server. Package must be .tar.gz or .rpm")).build();
+        }
+        try {
+            URL url = new URL(locationToFetchPackage);
+            new Thread(new PrestoAsynchronousCommand(new PrestoInstaller(installedPackageType, url, disableDependencyChecking))).start();
```
jirassimok: The first part of this line is used frequently enough that it should be its own function.

```java
runAsynchronously(PrestoCommand command)
{
    new Thread(() -> {
        try {
            new PrestoUpgrader(installedPackageType, url, disableDependencyChecking).runCommand();
        }
        catch (PrestoManagerException e) {
            // TODO: handle exception
        }
    });}
```
(This also eliminates the need for `PrestoAsynchronousCommand`.)

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -0,0 +1,146 @@
...
+    @PUT
+    @Consumes({"text/plain"})
+    @Produces({"text/plain"})
+    @ApiOperation(value = "Install presto using rpm or tarball")
+    @ApiResponses(value = {
+            @ApiResponse(code = 202, message = "Acknowledged request"),
+            @ApiResponse(code = 400, message = "Bad package format or Invalid url")
+    })
+    public synchronized Response install(@NotNull String locationToFetchPackage,
+            @QueryParam("disableDependencyChecking") @DefaultValue("false") boolean disableDependencyChecking)
```
jirassimok: `disableDependencyChecking` is a bit wordy for a query parameter.

I recommend renaming it to `checkDependencies` and making the default "true" (and either reversing it wherever it is used, or making the first line of the function read `boolean disableDependencyChecking = !checkDependencies`.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -0,0 +1,146 @@
...
+import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
+
+@Path("/package")
+@Api(description = "the package API")
+@Singleton
+public final class PackageAPI
+{
+    private static PackageType installedPackageType;
+
+    @PUT
```
jirassimok: This should be `POST`, not `PUT`, for requesting remote actions.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -0,0 +1,146 @@
...
+    @Consumes({"text/plain"})
+    @Produces({"text/plain"})
+    @ApiOperation(value = "Install presto using rpm or tarball")
+    @ApiResponses(value = {
+            @ApiResponse(code = 202, message = "Acknowledged request"),
+            @ApiResponse(code = 400, message = "Bad package format or Invalid url")
+    })
+    public synchronized Response install(@NotNull String locationToFetchPackage,
```
jirassimok: I do not believe that we should have `@NotNull` annotation here. It doesn't hurt, but it doesn't help significantly either, and we don't use it in any of the other API classes.

We shouldn't be calling this method, so there's little risk that the lack of documentation will cause problems.
If it's a functional constraint, it should be enforced in the program itself, via `requireNonNull`.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -0,0 +1,146 @@
...
+            this.command = command;
+        }
+
+        @Override
+        public void run()
+        {
+            try {
+                command.runCommand();
+            }
+            catch (Exception e) {
```
jirassimok: We almost certainly should not be catching _all_ exceptions here, especially not in this stage of development.

Only catch the exceptions you expect; other exceptions may indicate problems in the code.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -0,0 +1,146 @@
...
+            @ApiResponse(code = 400, message = "Bad package format or Invalid url")
+    })
+    public synchronized Response install(@NotNull String locationToFetchPackage,
+            @QueryParam("disableDependencyChecking") @DefaultValue("false") boolean disableDependencyChecking)
+    {
+        try {
+            installedPackageType = PackageType.valueOf(getFileExtension(locationToFetchPackage).toUpperCase());
+        }
+        catch (IllegalArgumentException e) {
+            return Response.status(BAD_REQUEST).entity(text("Invalid package type for presto-server. Package must be .tar.gz or .rpm")).build();
```
jirassimok: `Entity::text` is not required here; you can just pass the string in directly.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -0,0 +1,146 @@
...
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
+import javax.ws.rs.core.Response;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+
+import static com.google.common.io.Files.getFileExtension;
+import static javax.ws.rs.client.Entity.text;
```
jirassimok: This function's name is not clear enough on its own; `Entity` should be imported normally.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -0,0 +1,146 @@
...
+            return Response.status(BAD_REQUEST).entity(text("Invalid package type for presto-server. Package must be .tar.gz or .rpm")).build();
+        }
+        try {
+            final URL url = new URL(locationToFetchPackage);
+            new Thread(new PrestoAsynchronousCommand(new PrestoUpgrader(installedPackageType, url, disableDependencyChecking))).start();
+        }
+        catch (MalformedURLException e) {
+            return Response.status(BAD_REQUEST).entity(text("Invalid url")).build();
+        }
+        return Response.status(ACCEPTED).entity(text("Check status to make sure that presto is upgraded")).build();
```
jirassimok: The content of the response should be informational, not directional.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -0,0 +1,146 @@
...
+    @DELETE
+    @ApiOperation(value = "Uninstall presto")
+    @ApiResponses(value = {
+            @ApiResponse(code = 202, message = "Acknowledged request")
+    })
+    public synchronized Response uninstall(@QueryParam("disableDependencyChecking") @DefaultValue("false") boolean disableDependencyChecking,
+            @QueryParam("ignoreErrors") @DefaultValue("false") boolean ignoreErrors)
+    {
+        new Thread(new PrestoAsynchronousCommand(new PrestoUninstaller(installedPackageType, disableDependencyChecking, ignoreErrors))).start();
+        return Response.status(ACCEPTED).build();
```
jirassimok: This response should contain an informational message. Maybe "Request acknowledged; attempting to uninstall Presto".

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
sudarsanm94: I believe its not a good idea to use ternary operators. It makes the code difficult to understand.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
sudarsanm94: Isn't this same as using Runnable class?

jirassimok:
One more comment:

The lines in here get far too long for comfort; I have to scroll to view them on Github, and the scroll bar is only present at the bottom of each file...

I recommend wrapping the lines at 90 or so characters, with 120 characters as an absolute maximum (though Christina says 150 at longest, and `airlift/codestyle` says 180).

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
sudarsanm94: POST is used for upgrade. 
Generally PUT is good when we are creating something

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
jirassimok:
```java
runAsynchronously(PrestoCommand command)
{
    new Thread(new PrestoAsynchronousCommand(command)).start();
}
```

sudarsanm94: I thinks its a better idea to inline the method if it doesn't implement any significant logic.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
sudarsanm94: I'll update this

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
sudarsanm94: This is because run() method should not throw any exception

rschlussel: or more to the point you need to log anything weird that happens because otherwise you'll lose the error.

PR #14, opened 2017-07-10
Add API to access Presto logs

jirassimok:
# Important

If you can't add to a comment thread because the "reply" box disappears when clicked, find the comment thread in the "Files changed" tab, and reply there.

### DIFF
--- a/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentServer.java
+++ b/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentServer.java
@@ -14,6 +14,7 @@
 package com.teradata.prestomanager.agent;
 
 import com.teradata.prestomanager.agent.api.ConfigAPI;
+import com.teradata.prestomanager.agent.api.LogsAPI;
 import com.teradata.prestomanager.agent.api.PackageAPI;
 import com.teradata.prestomanager.common.ServerBuilder;
 import org.eclipse.jetty.server.Server;
@@ -32,7 +33,9 @@ class AgentServer
         Server server = new ServerBuilder()
                 .setURI(URI)
                 .setPort(PORT)
-                .registerClasses(ConfigAPI.class, PackageAPI.class)
+                .registerComponent(ConfigAPI.class)
+                .registerComponent(LogsAPI.class)
+                .registerComponent(PackageAPI.class)
                 .build();
 
         try {
--- a/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
+++ b/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
@@ -1,0 +1,281 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.teradata.prestomanager.agent;
+
+import com.google.common.collect.ImmutableMap;
+
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayDeque;
+import java.util.ConcurrentModificationException;
+import java.util.Deque;
+import java.util.Map;
+import java.util.StringJoiner;
+import java.util.function.Predicate;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Class for filtering the contents of log files
+ *
+ * Uses a regular expression to match log entries, and treats
+ * non-matching lines as part of the entry on the previous line.
+ *
+ * @see Builder
+ */
+public class LogFilter
+{
+    private final Path file;
+    private final Pattern logPattern;
+    private final String lineSeparator;
+    private final ImmutableMap<String, Predicate<String>> namedGroupFilters;
+    private final ImmutableMap<Integer, Predicate<String>> numberedGroupFilters;
+    private final int maxEntries;
+    private final boolean keepFirst;
+
+    private LogFilter(Path file, Pattern logPattern, String lineSeparator,
+            Map<String, Predicate<String>> namedGroupFilters,
+            Map<Integer, Predicate<String>> numberedGroupFilters,
+            int maxEntries, boolean keepFirst)
+            throws FileNotFoundException
+    {
+        this.file = requireNonNull(file);
+        this.logPattern = requireNonNull(logPattern);
+        this.lineSeparator = requireNonNull(lineSeparator);
+        this.namedGroupFilters = ImmutableMap.copyOf(requireNonNull(namedGroupFilters));
+        this.numberedGroupFilters = ImmutableMap.copyOf(requireNonNull(numberedGroupFilters));
+        this.maxEntries = maxEntries;
+        this.keepFirst = keepFirst;
+        if (!Files.exists(file)) {
+            throw new FileNotFoundException(file.toString());
+        }
+    }
+
+    public Stream<String> streamEntries()
+            throws IOException, ConcurrentModificationException
+    {
+        LogEntries logEntryBuilders = Files.lines(file).collect(
+                LogEntries::new,
+                LogEntries::addLine,
+                (a, b) -> {
+                    throw new ConcurrentModificationException();
+                });
+        return logEntryBuilders.streamEntries().map(StringJoiner::toString);
+    }
+
+    public static Builder builder()
+    {
+        return new Builder();
+    }
+
+    /**
+     * A builder for creating {@link LogFilter LogFilters}.
+     *
+     * Builder instances can be safely reused; {@link #build()} may be called
+     *  multiple times to build multiple filters.
+     */
+    public static class Builder
+    {
+        private Path file;
+        private Pattern logPattern;
+        private String lineSeparator = System.getProperty("line.separator");
+        private ImmutableMap.Builder<String, Predicate<String>> namedGroupFilters = ImmutableMap.builder();
+        private ImmutableMap.Builder<Integer, Predicate<String>> numberedGroupFilters = ImmutableMap.builder();
+        private int maxEntries = Integer.MAX_VALUE;
+        private boolean keepFirst = false;
+
+        private Builder() {}
+
+        public LogFilter build()
+                throws FileNotFoundException
+        {
+            return new LogFilter(file, logPattern, lineSeparator,
+                    namedGroupFilters.build(),
+                    numberedGroupFilters.build(),
+                    maxEntries, keepFirst);
+        }
+
+        public Builder setFile(Path file)
+        {
+            this.file = requireNonNull(file);
+            return this;
+        }
+
+        /**
+         * Use the given regular expression to match each line of a log file
+         */
+        public Builder setPattern(Pattern pattern)
+        {
+            this.logPattern = requireNonNull(pattern);
+            return this;
+        }
+
+        /**
+         * Add a filter to apply to a named capturing group
+         * in the regular expression for log file lines
+         */
+        public Builder addGroupFilter(String groupName, Predicate<String> filter)
+        {
+            namedGroupFilters.put(
+                    requireNonNull(groupName),
+                    requireNonNull(filter));
+            return this;
+        }
+
+        /**
+         * Add a filter to apply to a capturing group in
+         * the regular expression for log file lines
+         */
+        public Builder addGroupFilter(Integer groupNumber, Predicate<String> filter)
+        {
+            numberedGroupFilters.put(
+                    requireNonNull(groupNumber),
+                    requireNonNull(filter));
+            return this;
+        }
+
+        /**
+         * Use this string to separate lines in the same entry
+         */
+        public Builder setLineSeparator(String separator)
+        {
+            lineSeparator = requireNonNull(separator);
+            return this;
+        }
+
+        /**
+         * Set the maximum number of entries to keep when filtering.
+         */
+        public Builder setCapacity(int capacity)
+        {
+            maxEntries = capacity;
+            return this;
+        }
+
+        /**
+         * When there is a maximum capacity, keep the first entries
+         * instead of the last entries. Default false.
+         */
+        public Builder keepFirst(boolean shouldKeepFirst)
+        {
+            keepFirst = shouldKeepFirst;
+            return this;
+        }
+    }
+
+    /**
+     * Class representing a collection of log entries
+     */
+    private final class LogEntries
+    {
+        private Deque<StringJoiner> entries;
+        private Boolean keptLastLine;
+        private Matcher matcher;
+
+        private LogEntries()
+        {
+            entries = new ArrayDeque<>();
+            matcher = logPattern.matcher("");
+        }
+
+        Stream<StringJoiner> streamEntries()
+        {
+            return entries.stream();
+        }
+
+        /**
+         * Add a line from a log file to a deque of log entries
+         *
+         * If the string does not match {@link #logPattern}, it will be
+         * appended to the previous entry. If there is no previous entry, a new
+         * entry will be created only if entries are not being filtered.
+         *
+         * If {@link #maxEntries} is non-null and the deque contains that many
+         * entries, the oldest will be removed before the new entry is added.
+         */
+        void addLine(String line)
+        {
+            matcher.reset(line);
+            if (matcher.matches()) {
+                /* Apply all filters */
+                boolean passing = true;
+                for (ImmutableMap.Entry<String, Predicate<String>> f
+                        : namedGroupFilters.entrySet()) {
+                    passing = passing && f.getValue().test(matcher.group(f.getKey()));
+                }
+                for (ImmutableMap.Entry<Integer, Predicate<String>> f
+                        : numberedGroupFilters.entrySet()) {
+                    passing = passing && f.getValue().test(matcher.group(f.getKey()));
+                }
+
+                if (passing) {
+                    keptLastLine = addNewEntry(line);
+                }
+                else {
+                    keptLastLine = false;
+                }
+            }
+            else if (keptLastLine == null
+                    && namedGroupFilters.isEmpty() && numberedGroupFilters.isEmpty()) {
+                // If the first line does not match the pattern and there are no filters,
+                // use the line as a new entry.
+                keptLastLine = addNewEntry(line);
+            }
+            else if (keptLastLine != null && keptLastLine) {
+                keptLastLine = addToLast(line);
+            }
+            else {
+                keptLastLine = false;
+            }
+        }
+
+        /**
+         * @return true if the line addition succeeded
+         */
+        private boolean addNewEntry(String line)
+        {
+            if (entries.size() >= maxEntries) {
+                if (keepFirst) {
+                    return false;
+                }
+                else if (maxEntries == 0) {
+                    return true;
+                }
+                entries.removeFirst();
+            }
+            StringJoiner joiner = new StringJoiner(lineSeparator);
+            joiner.add(line);
+            entries.addLast(joiner);
+            return true;
+        }
+
+        /**
+         * @return true if the line addition succeeded
+         * (equivalently, true if any entries already exist)
+         */
+        private boolean addToLast(String line)
+        {
+            boolean entriesExist = !entries.isEmpty();
+            if (entriesExist) {
+                entries.getLast().add(line);
+            }
+            return entriesExist;
+        }
+    }
+}
--- a/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
+++ b/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
@@ -1,0 +1,248 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.teradata.prestomanager.agent;
+
+import com.teradata.prestomanager.common.JaxrsParameter;
+import org.apache.logging.log4j.Level;
+
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.Response.Status;
+
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardOpenOption;
+import java.time.Instant;
+import java.time.chrono.IsoChronology;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
+import java.time.format.DateTimeParseException;
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Utility class managing retrieval and deletion of Presto logs
+ */
+public final class LogsHandler
+{
+    // TODO: Make _all_ of these 'private static final' values configurable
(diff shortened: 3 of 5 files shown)


presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
```diff
@@ -0,0 +1,246 @@
...
+            .parseCaseInsensitive().parseStrict()
+            .append(DateTimeFormatter.ISO_LOCAL_DATE_TIME)
+            .appendPattern("Z")
+            .toFormatter().withChronology(IsoChronology.INSTANCE);
+
+    private static final String DATE_GROUP = "date";
+    private static final String LEVEL_GROUP = "level";
+    private static final Pattern LOG_REGEX = Pattern.compile(
+            "^(?<" + DATE_GROUP + ">[0-9]{4}-[0-9]{2}-[0-9]{2}T" +
+                    "[0-9]{2}:[0-9]{2}:[0-9]{2}(\\.[0-9]{3})?[+-][0-9]{4})" +
```
jirassimok: That's a literal dot in a regular expression, rather than a wildcard.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
```diff
@@ -0,0 +1,246 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static java.util.Objects.requireNonNull;
+import static javax.ws.rs.core.Response.Status;
+
+/**
+ * Class managing retrieval and deletion of Presto logs
+ */
+public class LogsHandler
```
jirassimok: Right, will do.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
```diff
@@ -0,0 +1,246 @@
...
+        requireNonNull(logLevel);
+
+        if (!startDate.isValid() || !endDate.isValid()) {
+            return Response.status(Status.BAD_REQUEST)
+                    .entity("Invalid date format").build();
+        }
+        Instant start = startDate.get();
+        Instant end = endDate.get();
+
+        if (start != null && end != null && maxEntries != null) {
```
jirassimok: That's a different value, see a couple lines above here. `start` and `end` will be null if they were not given as arguments.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
```diff
@@ -0,0 +1,246 @@
...
+        }
+
+        if (!logLevel.isValid()) {
+            return Response.status(Status.BAD_REQUEST)
+                    .entity("Invalid log level").build();
+        }
+        Level level = logLevel.get();
+
+        Path filePath = LOG_DIRECTORY.resolve(filename);
+        if (!Files.isRegularFile(filePath)) {
```
jirassimok: I'm just checking that it's not a directory or other special file.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
```diff
@@ -0,0 +1,246 @@
...
+            return Response.status(Status.NOT_FOUND)
+                    .entity("Invalid log file").build();
```
jirassimok: If it's a directory, it exists, but is not a log file.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
```diff
@@ -0,0 +1,246 @@
...
+     */
+    public static Response getLogs(String filename, JaxrsParameter<Instant> startDate,
+            JaxrsParameter<Instant> endDate, JaxrsParameter<Level> logLevel,
+            Integer maxEntries)
+    {
+        requireNonNull(startDate);
+        requireNonNull(endDate);
+        requireNonNull(logLevel);
+
+        if (!startDate.isValid() || !endDate.isValid()) {
```
jirassimok: `null` represents a parameter that was not given, not a parameter given a bad value.

But I even avoided that because of an apparent bug in Jersey that I previously encountered (though I don't think it would cause any problems in the current implementation).

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
```diff
@@ -0,0 +1,246 @@
...
+
+    private static Predicate<String> getFilter(Instant start, Instant end)
+    {
+        boolean nullStart = start == null;
+        boolean nullEnd = end == null;
+        if (nullStart && nullEnd) {
+            return null;
+        }
+        else if (nullStart) {
+            return s -> DATE_FORMAT.parse(s, Instant::from).compareTo(end) <= 0;
```
jirassimok: This is a simple function that consumes an argument `s` and returns the value of `DATE_FORMAT.parse(s, Instant::from).compareTo(end) <= 0`.

Because the first argument of `DateTimeFormatter::parse` is a `String`, the type of `s` is `String`.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
```diff
@@ -0,0 +1,286 @@
...
+    {
+        return new Builder();
+    }
+
+    /**
+     * A builder for creating {@link LogFilter LogFilters}.
+     *
+     * Builder instances can be safely reused.
+     */
+    public static class Builder
```
jirassimok: A single `requireNonNull` doesn't get in the way in a simple function (see e.g. `java.util.stream.StreamSupport::stream`).

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
ShanBai6: `\\` this is a backslash followed by a wildcard right?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
```diff
@@ -0,0 +1,286 @@
...
+         * appended to the previous entry. If there is no previous entry, a new
+         * entry will be created only if entries are not being filtered.
+         *
+         * If {@link #maxEntries} is non-null and the deque contains that many
+         * entries, the oldest will be removed before the new entry is added.
+         */
+        void addLine(String line)
+        {
+            Matcher matcher = logPattern.matcher(line);
+            if (matcher.matches()) {
```
jirassimok: The log will always follow a specific format, which should be configurable.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
ShanBai6: So it return a boolean?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
```diff
@@ -0,0 +1,286 @@
...
+                /* Apply all filters */
+                boolean passing = true;
+                for (ImmutableMap.Entry<String, Predicate<String>> f
+                        : namedGroupFilters.entrySet()) {
+                    passing = passing && f.getValue().test(matcher.group(f.getKey()));
```
jirassimok: True, but the cost of testing each predicate is considerably greater than the cost of a single boolean check, and `&&` allows short-circuiting (of course, with only two predicates actually in use, the actual cost is negligible either way).

Though `&=` does look nicer. I'll use whichever you recommend after reading this.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
```diff
@@ -0,0 +1,286 @@
...
+                }
+                else {
+                    keptLastLine = false;
+                }
+            }
+            else if (keptLastLine == null
+                    && namedGroupFilters.isEmpty() && numberedGroupFilters.isEmpty()) {
+                // If the first line does not match the pattern and there are no filters,
+                // use the line as a new entry.
+                keptLastLine = addNewEntry(line);
```
jirassimok: Correct.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
ShanBai6: I think at this point, `startDate` and `endDate` will be both valid & non-null. How could get return a null value

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
ShanBai6: If it is a txt file, you will just read it then? I think you should check explicitly for a log file

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
ShanBai6: Here I am saying provide a different message for not found.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
```diff
@@ -0,0 +1,286 @@
...
+            return true;
+        }
+
+        /**
+         * @return true if the line addition succeeded
+         * (equivalently, true if any entries already exist)
+         */
+        private boolean addToLast(String line)
+        {
+            boolean isNotEmpty = !entries.isEmpty();
```
jirassimok: It seems a bit more readable to me this way. (Though I've just renamed the variable to `entriesExist`, which makes it a bit clearer.)

pom.xml
```diff
@@ -33,6 +33,7 @@
         <!-- Dependency versions -->
         <jetty.version>9.4.6.v20170531</jetty.version>
         <jersey.version>2.26-b06</jersey.version>
+        <log4j.version>2.8.2</log4j.version>
```
ShanBai6: This is just a reminder

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
jirassimok: One slash to escape the backslash for Java, leaving one behind for the regex. So `\\.` becomes `\.` in the regex.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
jirassimok: `JaxrsParameters` are essentially wrappers around the values given as parameters by Jersey/JAX-RS.

If the implementation of `parseString` returns null for some values, then `get()` will return null as well. The implementation in `DateParameter` returns null if the parameter is "DEFAULT", which is used as the default value whenever `DateParameter` is used.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
jirassimok: A log file isn't necessarily named "something.log", so I haven't checked for that; if it's a file in the log directory and it has the name you gave, it can be read.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
jirassimok: Oh, that's a good idea.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
ShanBai6: `12:00:00.000` should be the format, why should there be a backslash in the middle like `12:00:00\.000`

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
jirassimok: `getFilter` returns a `Predicate<String>`, which represents a function that consumes a String and returns a boolean.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
jirassimok: `.` is the wildcard in a regex.

`Pattern.matches("0.0", "000")` (regex `0.0`) returns true.
`Pattern.matches("0\\.0", "000")` (regex `0\.0`) returns false.

pom.xml
sudarsanm94: I think it's  better to update your master branch and rebase your logs-api branch on top of it.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentServer.java
```diff
@@ -32,6 +33,7 @@ public static void main(String[] args)
                 .setURI(URI)
                 .setPort(PORT)
                 .registerComponent(ConfigAPI.class)
```
sudarsanm94: You can use registerClasses here

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
jirassimok: test

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentServer.java
jirassimok: They have the exact same result. I prefer the separate calls because then to change the included classes, any of the classes' lines can be removed, without worrying about the trailing parenthesis or commas.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
```diff
@@ -0,0 +1,288 @@
...
+    private final Path file;
+    private final Pattern logPattern;
+    private final String lineSeparator;
+    private final ImmutableMap<String, Predicate<String>> namedGroupFilters;
+    private final ImmutableMap<Integer, Predicate<String>> numberedGroupFilters;
+    private final Integer maxEntries;
+    private final boolean keepFirst;
+
+    private LogFilter(Path file, Pattern logPattern, String lineSeparator,
+            ImmutableMap<String, Predicate<String>> namedGroupFilters,
```
maciejgrzybek: Please use regular `Map` instead of `ImmutableMap`, then you don't have to introduce extra dependency for someone using that class while if the underlying `Map` in fact is `ImmutableMap`, then it won't be copied anyway (the same as if you used `ImmutableMap` explicitly), so there's no benefit of requiring the `Map` to be of type `ImmutableMap`.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
```diff
@@ -0,0 +1,288 @@
...
+            ImmutableMap<Integer, Predicate<String>> numberedGroupFilters,
+            Integer maxEntries, boolean keepFirst)
+            throws FileNotFoundException
+    {
+        this.file = requireNonNull(file);
+        this.logPattern = requireNonNull(logPattern);
+        this.lineSeparator = requireNonNull(lineSeparator);
+        this.namedGroupFilters = requireNonNull(namedGroupFilters);
+        this.numberedGroupFilters = requireNonNull(numberedGroupFilters);
+        this.maxEntries = maxEntries; // null -> no maximum
```
maciejgrzybek: Why not `Integer.MAX_VALUE` instead?
Null does not fit nicely in the arithmetics and requires special ifology wherever you use the value.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
```diff
@@ -0,0 +1,288 @@
...
+    private LogFilter(Path file, Pattern logPattern, String lineSeparator,
+            ImmutableMap<String, Predicate<String>> namedGroupFilters,
+            ImmutableMap<Integer, Predicate<String>> numberedGroupFilters,
+            Integer maxEntries, boolean keepFirst)
+            throws FileNotFoundException
+    {
+        this.file = requireNonNull(file);
+        this.logPattern = requireNonNull(logPattern);
+        this.lineSeparator = requireNonNull(lineSeparator);
+        this.namedGroupFilters = requireNonNull(namedGroupFilters);
```
maciejgrzybek: Use `ImmutableMap.copyOf()`, then you don't need to require `ImmutableMap`, regular `Map` would work the same good here.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
```diff
@@ -0,0 +1,288 @@
...
+
+    public static Builder builder()
+    {
+        return new Builder();
+    }
+
+    /**
+     * A builder for creating {@link LogFilter LogFilters}.
+     *
+     * Builder instances can be safely reused.
```
maciejgrzybek: Did you mean it's thread safe or reentrant?
If so, please annotate the class appropriately, e.g. with `@javax.annotation.concurrent.ThreadSafe`.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
```diff
@@ -0,0 +1,288 @@
...
+            return new LogFilter(file, logPattern, lineSeparator,
+                    namedGroupFilters.build(),
+                    numberedGroupFilters.build(),
+                    maxEntries, keepFirst);
+        }
+
+        public Builder setFile(Path file)
+        {
+            requireNonNull(file);
+            this.file = file;
```
maciejgrzybek: I think in the rest of code you used to use `this.X = requireNonNull(x, "foo")` pattern.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
```diff
@@ -0,0 +1,288 @@
...
+            return this;
+        }
+
+        /**
+         * Use the given regular expression to match each line of a log file
+         */
+        public Builder setPattern(Pattern pattern)
+        {
+            requireNonNull(pattern);
+            this.logPattern = pattern;
```
maciejgrzybek: ditto

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
```diff
@@ -0,0 +1,288 @@
...
+        }
+
+        /**
+         * Add a filter to apply to a named capturing group
+         * in the regular expression for log file lines
+         */
+        public Builder addGroupFilter(String groupName, Predicate<String> filter)
+        {
+            requireNonNull(groupName);
+            if (filter != null) {
```
maciejgrzybek: Why are you purposefully allowing `null` filter?
Why one would `addGroupFilter` without any filter?
If one want to allow any string, he'd use `(s) -> true`, I believe, or better - not adding such group filter to the builder at all.
Or if you feel strongly about adding empty filters, please express it using `Optional<Predicate<String>>` so you avoid the ambiguity and presumed implicit logic in `Builder`.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
```diff
@@ -0,0 +1,288 @@
...
+        }
+
+        /**
+         * Add a filter to apply to a capturing group in
+         * the regular expression for log file lines
+         */
+        public Builder addGroupFilter(Integer groupNumber, Predicate<String> filter)
+        {
+            requireNonNull(groupNumber);
+            if (filter != null) {
```
maciejgrzybek: ditto

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
```diff
@@ -0,0 +1,288 @@
...
+            return this;
+        }
+
+        /**
+         * Use this string to separate lines in the same entry
+         */
+        public Builder setLineSeparator(String separator)
+        {
+            requireNonNull(separator);
+            lineSeparator = separator;
```
maciejgrzybek: ditto

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
```diff
@@ -0,0 +1,288 @@
...
+            return this;
+        }
+
+        /**
+         * Set the maximum number of entries to keep when filtering.
+         *
+         * A null capacity results in all entries being kept.
```
maciejgrzybek: Use `Optional` instead of `null`, to avoid ambiguity or add an extra check for nullness (`requireNonNull`) because if one doesn't want to set capacity, he should not call this method.
If you want to allow changing capacity before `build()` is called, add a `unlimitedCapacity()` method to zero the effect of previously set capacity.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
```diff
@@ -0,0 +1,288 @@
...
+        {
+            maxEntries = capacity;
+            return this;
+        }
+
+        /**
+         * When there is a maximum capacity, keep the first entries
+         * instead of the last entries
+         */
+        public Builder keepFirst()
```
maciejgrzybek: Maybe consider adding `unkeepFirst` method as well (or simply `setKeepFirst(boolean keepFirst)`) in case you want to allow `Builder` changing the behavior, e.g.
```
Builder builder;
builder.keepFirst();
if (...) {
builder.unKeepFirst();
}
```

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
jirassimok: Done.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
jirassimok: If a maximum is provided, I allocate a Deque with that much space. A Deque with MAX_VALUE elements would not work out well.

But I suppose probably not the best idea to pre-allocate the Deque like that anyway, in case someone manually gives a significantly large number. I'll change that.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
jirassimok: The builder creates new contents for the filter, so `build()` may safely be used repeatedly.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
jirassimok: Done.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
jirassimok: Fixed.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
jirassimok: Because a non-present parameter will be represented by `null` in `LogsHandler`, and if I ignore `null` filters, then I don't have to do a null check when adding them.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
jirassimok: Fixed.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
jirassimok: Fixed.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
jirassimok: I'll add `keepFirst(boolean)`.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
ShanBai6: Second part does take a little more time, but I guess performance at this point does not matter that much, up to you.

jirassimok:
@maciejgrzybek I've addressed your comments.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
```diff
@@ -0,0 +1,248 @@
...
+            };
+        }
+    }
+
+    private static Predicate<String> getFilter(Level level)
+    {
+        if (level == null || level == LOG_ALL) {
+            return s -> true;
+        }
+        else {
```
maciejgrzybek: Check style comment:
you don't need `else` here.

PR #15, opened 2017-07-11
Add control API for agent

sudarsanm94:
Add utility methods to start, stop and restart presto; Get presto status

### DIFF
--- a/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoRpmControlUtils.java
+++ b/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoRpmControlUtils.java
@@ -1,0 +1,259 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.teradata.prestomanager.agent;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.glassfish.jersey.client.JerseyClient;
+
+import javax.ws.rs.ProcessingException;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.UriBuilder;
+
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+import static com.teradata.prestomanager.agent.AgentFileUtils.getFileProperty;
+import static com.teradata.prestomanager.agent.CommandExecutor.executeCommand;
+import static java.lang.String.format;
+import static javax.ws.rs.client.Entity.entity;
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+import static javax.ws.rs.core.MediaType.TEXT_PLAIN;
+import static javax.ws.rs.core.Response.Status.ACCEPTED;
+import static javax.ws.rs.core.Response.Status.CONFLICT;
+import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
+import static javax.ws.rs.core.Response.Status.NOT_FOUND;
+import static javax.ws.rs.core.Response.Status.OK;
+import static javax.ws.rs.core.UriBuilder.fromUri;
+import static org.glassfish.jersey.client.JerseyClientBuilder.createClient;
+
+public final class PrestoRpmControlUtils
+{
+    private static final Logger LOGGER = LogManager.getLogger(PrestoRpmControlUtils.class);
+    private static final JerseyClient JERSEY_CLIENT = createClient();
+    private static final String LAUNCHER_SCRIPT = "/usr/lib/presto/bin/launcher";
+    private static final Path PRESTO_CONFIG_FILE = Paths.get("/etc/presto/config.properties");
+    private static final Gson GSON = new Gson();
+    private static final int SUBPROCESS_TIMEOUT = 120;
+
+    private PrestoRpmControlUtils() {}
+
+    public static Response startUsingRpm()
+    {
+        try {
+            if (!isInstalled()) {
+                LOGGER.error("Presto is not installed");
+                return Response.status(NOT_FOUND).entity("Presto is not installed").build();
+            }
+        }
+        catch (PrestoManagerException e) {
+            LOGGER.error(e.getMessage(), e.getCause());
+            return Response.status(INTERNAL_SERVER_ERROR).build();
+        }
+        new Thread(() -> {
+            try {
+                int startPresto = executeCommand(SUBPROCESS_TIMEOUT, "service", "presto", "start");
+                if (startPresto != 0) {
+                    LOGGER.error("Failed to start presto");
+                }
+            }
+            catch (PrestoManagerException e) {
+                LOGGER.error(e.getMessage(), e.getCause());
+            }
+        }).start();
+        return Response.status(ACCEPTED).entity("Presto is being started. " +
+                "Please check back later using the status api to make sure that Presto has successfully started.").build();
+    }
+
+    public static Response stopUsingRpm(StopType stopType)
+    {
+        try {
+            if (!isInstalled()) {
+                LOGGER.error("Presto is not installed");
+                return Response.status(NOT_FOUND).build();
+            }
+            if (!isRunning()) {
+                LOGGER.info("Presto is not running");
+                return Response.status(OK).build();
+            }
+            switch (stopType) {
+                case TERMINATE:
+                    terminatePresto();
+                    break;
+                case KILL:
+                    killPresto();
+                    break;
+                case GRACEFUL:
+                    if (isCoordinator()) {
+                        LOGGER.error("Coordinator can't be gracefully stopped.");
+                        return Response.status(CONFLICT).build();
+                    }
+                    gracefulStop();
+                    break;
+                default:
+                    LOGGER.error("Invalid stop type: {}", stopType);
+                    return Response.status(INTERNAL_SERVER_ERROR).build();
+            }
+        }
+        catch (PrestoManagerException e) {
+            LOGGER.error(e.getMessage(), e.getCause());
+            return Response.status(INTERNAL_SERVER_ERROR).build();
+        }
+        return Response.status(OK).build();
+    }
+
+    public static Response restartUsingRpm()
+    {
+        try {
+            if (!isInstalled()) {
+                LOGGER.error("Presto is not installed");
+                return Response.status(NOT_FOUND).build();
+            }
+            int prestoRestart = executeCommand(SUBPROCESS_TIMEOUT, "service", "presto", "restart");
+            if (prestoRestart != 0) {
+                throw new PrestoManagerException("Failed to restart presto", prestoRestart);
+            }
+        }
+        catch (PrestoManagerException e) {
+            LOGGER.error(e.getMessage(), e.getCause());
+            return Response.status(INTERNAL_SERVER_ERROR).build();
+        }
+        return Response.status(OK).build();
+    }
+
+    /**
+     * The status api can be used to ascertain whether presto is installed and running.
+     * If Presto is running, this method would return a JSON file containing the following details
+     * nodeVersion : Presto version installed in the node
+     * environment : The name of the environment
+     * coordinator : Specifies whether the current node functions as a coordinator
+     * state : The state of the node. ACTIVE (or) SHUTTING_DOWN
+     *
+     * @return Response containing node status
+     */
+    public static Response statusUsingRpm()
+    {
+        try {
+            if (!isInstalled()) {
+                LOGGER.info("Presto is not installed");
+                return Response.status(OK).entity(GSON.toJson("Presto is not installed"))
+                        .type(APPLICATION_JSON).build();
+            }
+        }
+        catch (PrestoManagerException e) {
+            LOGGER.error(e.getMessage(), e.getCause());
+            return Response.status(INTERNAL_SERVER_ERROR).build();
+        }
+        try {
+            String prestoPort = getPrestoPort();
+            UriBuilder uriBuilder = fromUri(format("http://localhost:%s", prestoPort)).path("/v1/info");
+            Response prestoInfo = JERSEY_CLIENT.target(uriBuilder.build()).request(APPLICATION_JSON).buildGet().invoke();
+            Response prestoState = JERSEY_CLIENT.target(uriBuilder.path("/state").build())
+                    .request(APPLICATION_JSON).buildGet().invoke();
+            JsonParser jsonParser = new JsonParser();
+            JsonObject prestoStatus = jsonParser.parse(prestoInfo.readEntity(String.class)).getAsJsonObject();
+            prestoStatus.addProperty("state", jsonParser.parse(prestoState.readEntity(String.class)).getAsString());
+            return Response.status(OK).entity(prestoStatus.toString()).type(APPLICATION_JSON).build();
+        }
+        catch (ProcessingException e) {
+            LOGGER.info("Presto is not running");
+            return Response.status(OK).entity(GSON.toJson("Presto is installed but not running"))
+                    .type(APPLICATION_JSON).build();
+        }
+        catch (IOException e) {
+            LOGGER.error("Failed to get status.", e);
+            return Response.status(INTERNAL_SERVER_ERROR).entity(GSON.toJson("Failed to get status"))
+                    .type(APPLICATION_JSON).build();
+        }
+    }
+
+    private static boolean isInstalled()
+            throws PrestoManagerException
+    {
+        return executeCommand("rpm", "-q", "presto-server-rpm") == 0;
+    }
+
+    private static boolean isRunning()
+            throws PrestoManagerException
+    {
+        return isInstalled() && executeCommand("service", "presto", "status") == 0;
+    }
+
+    private static void terminatePresto()
+            throws PrestoManagerException
+    {
+        int prestoTerminate = executeCommand("service", "presto", "stop");
+        if (prestoTerminate != 0) {
+            throw new PrestoManagerException("Failed to stop presto", prestoTerminate);
+        }
+    }
+
+    private static void killPresto()
+            throws PrestoManagerException
+    {
+        executeCommand("sudo", LAUNCHER_SCRIPT, "kill");
+        int prestoKill = executeCommand("service", "presto", "status");
+        if (prestoKill != 3) {
+            throw new PrestoManagerException("Failed to stop presto", prestoKill);
+        }
+    }
+
+    private static void gracefulStop()
+            throws PrestoManagerException
+    {
+        try {
+            String prestoPort = getPrestoPort();
+            UriBuilder uriBuilder = fromUri(format("http://localhost:%s", prestoPort)).path("/v1/info/state");
+            Response response = JERSEY_CLIENT.target(uriBuilder.build()).request(TEXT_PLAIN)
+                    .buildPut(entity(GSON.toJson("SHUTTING_DOWN"), APPLICATION_JSON)).invoke();
+            if (response.getStatus() != 200) {
+                throw new PrestoManagerException("Failed to stop presto gracefully");
+            }
+        }
+        catch (ProcessingException e) {
+            LOGGER.info("Presto is not running", e);
+        }
+        catch (IOException e) {
+            throw new PrestoManagerException("Failed to get Presto port number.", e);
+        }
+    }
+
+    private static String getPrestoPort()
+            throws IOException
+    {
+        return getFileProperty(PRESTO_CONFIG_FILE, "http-server.http.port");
+    }
+
+    private static boolean isCoordinator()
+            throws PrestoManagerException
+    {
+        try {
+            String prestoPort = getPrestoPort();
+            UriBuilder uriBuilder = fromUri(format("http://localhost:%s", prestoPort)).path("/v1/info/coordinator");
+            Response isCoordinator = JERSEY_CLIENT.target(uriBuilder.build()).request(TEXT_PLAIN).buildGet().invoke();
+            return isCoordinator.getStatus() == 200;
+        }
+        catch (ProcessingException e) {
+            LOGGER.info("Presto is not running", e);
+        }
+        catch (IOException e) {
+            throw new PrestoManagerException("Failed to get Presto port number.", e);
+        }
+        return false;
+    }
+}
--- a/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/ControlAPI.java
+++ b/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/ControlAPI.java
@@ -1,0 +1,106 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.teradata.prestomanager.agent.api;
+
+import com.teradata.prestomanager.agent.StopType;
+import io.swagger.annotations.Api;
+import io.swagger.annotations.ApiOperation;
+import io.swagger.annotations.ApiParam;
+import io.swagger.annotations.ApiResponse;
+import io.swagger.annotations.ApiResponses;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import javax.inject.Singleton;
+import javax.ws.rs.DefaultValue;
+import javax.ws.rs.GET;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
+import javax.ws.rs.core.Response;
+
+import static com.teradata.prestomanager.agent.PrestoRpmControlUtils.restartUsingRpm;
+import static com.teradata.prestomanager.agent.PrestoRpmControlUtils.startUsingRpm;
+import static com.teradata.prestomanager.agent.PrestoRpmControlUtils.statusUsingRpm;
+import static com.teradata.prestomanager.agent.PrestoRpmControlUtils.stopUsingRpm;
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+import static javax.ws.rs.core.MediaType.TEXT_PLAIN;
+
+@Path("/presto")
+@Api(description = "API to start, stop, restart presto and get presto status")
+@Singleton
+public class ControlAPI
+{
+    private static final Logger LOGGER = LogManager.getLogger(ControlAPI.class);
+
+    @POST
+    @Path("/start")
+    @Produces(TEXT_PLAIN)
+    @ApiOperation(value = "Start presto")
+    @ApiResponses(value = {
+            @ApiResponse(code = 202, message = "Acknowledged request"),
+            @ApiResponse(code = 404, message = "Presto is not installed"),
+            @ApiResponse(code = 500, message = "Failed to start presto")
+    })
+    public synchronized Response startPresto()
+    {
+        LOGGER.debug("POST /presto/start");
+        return startUsingRpm();
+    }
+
+    @POST
+    @Path("/stop")
+    @ApiOperation(value = "Stop presto")
+    @ApiResponses(value = {
+            @ApiResponse(code = 200, message = "Successfully stopped presto"),
+            @ApiResponse(code = 404, message = "Presto is not installed"),
+            @ApiResponse(code = 409, message = "Coordinator can't be gracefully stopped"),
+            @ApiResponse(code = 500, message = "Failed to stop presto")
+    })
+    public synchronized Response stopPresto(@QueryParam("stopType") @ApiParam("StopType: TERMINATE, KILL or GRACEFUL")
+    @DefaultValue("GRACEFUL") StopType stopType)
+    {
+        LOGGER.debug("POST /presto/stop ; stopType: {}", stopType);
+        return stopUsingRpm(stopType);
+    }
+
+    @POST
+    @Path("/restart")
+    @ApiOperation(value = "Restart presto")
+    @ApiResponses(value = {
+            @ApiResponse(code = 200, message = "Successfully restarted presto"),
+            @ApiResponse(code = 404, message = "Presto is not installed"),
+            @ApiResponse(code = 500, message = "Failed to restart presto")
+    })
+    public synchronized Response restartPresto()
+    {
+        LOGGER.debug("POST /presto/restart");
+        return restartUsingRpm();
+    }
+
+    @GET
+    @Path("/status")
+    @Produces(APPLICATION_JSON)
+    @ApiOperation(value = "Get presto status")
+    @ApiResponses(value = {
+            @ApiResponse(code = 200, message = "Successfully retrieved presto status"),
+            @ApiResponse(code = 500, message = "Failed to get presto status")
+    })
+    public synchronized Response prestoStatus()
+    {
+        LOGGER.debug("GET /presto/status");
+        return statusUsingRpm();
+    }
+}
--- a/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentServer.java
+++ b/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentServer.java
@@ -15,6 +15,7 @@ package com.teradata.prestomanager.agent;
 
 import com.teradata.prestomanager.agent.api.ConfigAPI;
 import com.teradata.prestomanager.agent.api.ConnectorsAPI;
+import com.teradata.prestomanager.agent.api.ControlAPI;
 import com.teradata.prestomanager.agent.api.LogsAPI;
 import com.teradata.prestomanager.agent.api.PackageAPI;
 import com.teradata.prestomanager.common.ServerBuilder;
@@ -38,6 +39,7 @@ class AgentServer
                 .registerComponent(ConnectorsAPI.class)
                 .registerComponent(LogsAPI.class)
                 .registerComponent(PackageAPI.class)
+                .registerComponent(ControlAPI.class)
                 .build();
 
         try {
--- a/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/StopType.java
+++ b/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/StopType.java
@@ -1,0 +1,19 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.teradata.prestomanager.agent;
+
+public enum StopType
+{
+    GRACEFUL, TERMINATE, KILL
+}


presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
```diff
@@ -0,0 +1,146 @@
...
+
+import static com.google.common.util.concurrent.Uninterruptibles.sleepUninterruptibly;
+import static com.teradata.prestomanager.agent.AgentFileUtils.getFileProperty;
+import static com.teradata.prestomanager.agent.CommandExecutor.executeCommand;
+import static com.teradata.prestomanager.common.RequestBuilder.newRequestBuilder;
+import static java.lang.String.format;
+import static javax.ws.rs.client.Entity.json;
+import static javax.ws.rs.core.Response.Status.OK;
+
+// TODO: stop presto before upgrade and uninstall
```
ShanBai6: You would just call the stop method in your other class, not a todo here

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
```diff
@@ -0,0 +1,146 @@
...
+public final class AgentStatusUtils
+{
+    private static final long WAIT_TIME_SECONDS = 60;
+    private static final long STEP_MILLISECONDS = 100;
```
ShanBai6: If number are small, just use int

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
```diff
@@ -0,0 +1,146 @@
...
+    private static final String LAUNCHER_SCRIPT = "/usr/lib/presto/bin/launcher";
```
ShanBai6: This could be path as well.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
```diff
@@ -0,0 +1,146 @@
...
+            throw new PrestoManagerException("Presto is not running");
+        }
+        switch (stopType) {
+            case TERMINATE:
+                if (executeCommand("service presto stop") != 0) {
+                    throw new PrestoManagerException("Failed to stop presto");
+                }
+                break;
+            case KILL:
+                executeCommand(format("sudo %s kill", LAUNCHER_SCRIPT));
```
ShanBai6: Does this need sudo?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
```diff
@@ -0,0 +1,146 @@
...
+        if (executeCommand("service presto start") != 0) {
+            throw new PrestoManagerException("Failed to start presto");
+        }
+        long endTime = System.currentTimeMillis() + TimeUnit.SECONDS.toMillis(WAIT_TIME_SECONDS);
+        while (System.currentTimeMillis() < endTime && !Thread.currentThread().isInterrupted()) {
+            if (executeCommand("service presto status") == 0) {
+                return;
+            }
+            else {
+                sleepUninterruptibly(STEP_MILLISECONDS, TimeUnit.MILLISECONDS);
```
ShanBai6: Make sure the values are proper

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
```diff
@@ -0,0 +1,146 @@
...
+
+    public static void stopPresto(StopType stopType)
+            throws PrestoManagerException
+    {
+        if (executeCommand("service presto status") != 0) {
+            throw new PrestoManagerException("Presto is not running");
+        }
+        switch (stopType) {
+            case TERMINATE:
+                if (executeCommand("service presto stop") != 0) {
```
ShanBai6: I would just make sure it is not started, then return. I won't throw the exception, make the command idempotent.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
```diff
@@ -0,0 +1,146 @@
...
+    private AgentStatusUtils() {}
+
+    public static void startPresto()
+            throws PrestoManagerException
+    {
+        if (executeCommand("rpm -q presto-server-rpm") != 0) {
+            throw new PrestoManagerException(format("Presto not installed"));
+        }
+        if (executeCommand("service presto status") == 0) {
+            throw new PrestoManagerException("Presto is already running");
```
ShanBai6: I would not throw exception here, just return.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
```diff
@@ -0,0 +1,146 @@
...
+                        throw new PrestoManagerException("Coordinator can't be gracefully stopped");
+                    }
+                    String prestoPort = getFileProperty(PATH_TO_CONFIG_FILE, "http-server.http.port");
+                    Response response = newRequestBuilder(new URI(format("localhost:%s/v1/info/state", prestoPort)))
+                            .withBody(json("SHUTTING_DOWN")).buildPut().invoke();
+                    if (response.getStatus() != 200) {
+                        throw new PrestoManagerException("Failed to stop presto gracefully");
+                    }
+                }
+                catch (IOException | URISyntaxException e) {
```
ShanBai6: Since you are catching this exception, I would put this URI as a parameter of the method.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
```diff
@@ -0,0 +1,146 @@
...
+
+    public static void restartPresto()
+            throws PrestoManagerException
+    {
+        if (executeCommand("rpm -q presto-server-rpm") != 0) {
+            throw new PrestoManagerException(format("Presto not installed"));
+        }
+        if (executeCommand("service presto restart") != 0) {
+            throw new PrestoManagerException("Failed to restart presto");
+        }
```
ShanBai6: Don't you check here if it is already started as you did above?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
```diff
@@ -0,0 +1,146 @@
...
+    {
+        GRACEFUL, TERMINATE, KILL;
+
+        public static StopType fromString(String name)
+        {
+            try {
+                return valueOf(name.toUpperCase());
+            }
+            catch (IllegalArgumentException e) {
+                return GRACEFUL;
```
ShanBai6: Just throw a new PMinstall exception

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
```diff
@@ -0,0 +1,146 @@
...
+    {
+        if (executeCommand("rpm -q presto-server-rpm") != 0) {
+            return Response.status(OK).entity(json("Presto not installed")).build();
+        }
+        if (executeCommand("service presto status") != 0) {
+            return Response.status(OK).entity(json("Presto is installed but not running")).build();
+        }
+        try {
+            String prestoPort = getFileProperty(PATH_TO_CONFIG_FILE, "http-server.http.port");
+            return newRequestBuilder(new URI(format("localhost:%s/v1/info", prestoPort))).buildGet().invoke();
```
ShanBai6: Same as above, make this URI as parameter

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
jirassimok: I would not implement this TODO. Rather, if Presto is running when one of these operations is requested, the request should return a 409 ("Conflict") indicating that Presto is running.

The less magic behavior, the better, I say.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
jirassimok: What happens if you call `service presto start` while Presto is already running?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
sudarsanm94: Yes.
Since the launcher script is in /usr/..

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
sudarsanm94: Nothing, it'll just return already running

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
```diff
@@ -0,0 +1,146 @@
...
+        if (executeCommand("rpm -q presto-server-rpm") != 0) {
+            throw new PrestoManagerException(format("Presto not installed"));
+        }
+        if (executeCommand("service presto status") == 0) {
+            throw new PrestoManagerException("Presto is already running");
+        }
+        if (executeCommand("service presto start") != 0) {
+            throw new PrestoManagerException("Failed to start presto");
+        }
+        long endTime = System.currentTimeMillis() + TimeUnit.SECONDS.toMillis(WAIT_TIME_SECONDS);
```
jirassimok: I haven't looked it over too carefully, but this whole section looks bad to me. I'll expand on this comment after I finish the rest of the review.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
```diff
@@ -0,0 +1,146 @@
...
+import java.nio.file.Paths;
+import java.util.concurrent.TimeUnit;
+
+import static com.google.common.util.concurrent.Uninterruptibles.sleepUninterruptibly;
+import static com.teradata.prestomanager.agent.AgentFileUtils.getFileProperty;
+import static com.teradata.prestomanager.agent.CommandExecutor.executeCommand;
+import static com.teradata.prestomanager.common.RequestBuilder.newRequestBuilder;
+import static java.lang.String.format;
+import static javax.ws.rs.client.Entity.json;
+import static javax.ws.rs.core.Response.Status.OK;
```
jirassimok: This is too many static imports.

`Response.Status`, `RequestBuilder`, `Entity`, and `AgentFileUtils` should be imported normally (the last mainly because `getFileProperty` isn't particularly clear when used out of context).

`format`, `executeCommand`, and `sleepUninterruptibly` are fine.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
```diff
@@ -0,0 +1,146 @@
...
+                executeCommand(format("sudo %s kill", LAUNCHER_SCRIPT));
+                if (executeCommand("service presto status") != 3) {
+                    throw new PrestoManagerException("Failed to stop presto");
+                }
+                break;
+            case GRACEFUL:
+                try {
+                    if (getFileProperty(PATH_TO_CONFIG_FILE, "coordinator").equalsIgnoreCase("true")) {
+                        throw new PrestoManagerException("Coordinator can't be gracefully stopped");
+                    }
```
jirassimok: We need to determine a proper behavior for a graceful stop request for the coordinator; it shouldn't throw an exception for a normal response.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
```diff
@@ -0,0 +1,146 @@
...
+                    String prestoPort = getFileProperty(PATH_TO_CONFIG_FILE, "http-server.http.port");
```
jirassimok: This should be a field of the class for now, and configurable once we get a configuration framework running.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
```diff
@@ -0,0 +1,146 @@
...
+                    Response response = newRequestBuilder(new URI(format("localhost:%s/v1/info/state", prestoPort)))
```
jirassimok: This URI should be a field of the class, and should be validated before attempting to send to it, if possible.
(This whole section will need some shuffling once we have working configuration.)

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
```diff
@@ -0,0 +1,146 @@
...
+        try {
+            String prestoPort = getFileProperty(PATH_TO_CONFIG_FILE, "http-server.http.port");
+            return newRequestBuilder(new URI(format("localhost:%s/v1/info", prestoPort))).buildGet().invoke();
+        }
+        catch (IOException | URISyntaxException e) {
+            throw new PrestoManagerException("Failed to get status");
+        }
+    }
+
+    public enum StopType
```
jirassimok: This should be replaced with a `JaxrsParameter` (as seen in the logs API pull request), or with a `ParamConverter`/`ParamConverterProvider` once we figure out how to use those.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
sudarsanm94: Updated

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
sudarsanm94: This I use it withing format(). That's why its string.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
sudarsanm94: The `service presto restart` will take care of that

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
sudarsanm94: This is to make GRACEFUL as default if invalid arguments are specified

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
sudarsanm94: This part of the code is to periodically check whether presto is installed or not.
There's no standard lib to do this stuff.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
```diff
@@ -0,0 +1,146 @@
...
+    private static final long STEP_MILLISECONDS = 100;
+    private static final String LAUNCHER_SCRIPT = "/usr/lib/presto/bin/launcher";
+    private static final Path PATH_TO_CONFIG_FILE = Paths.get("/etc/presto/config.properties");
+
+    private AgentStatusUtils() {}
+
+    public static void startPresto()
+            throws PrestoManagerException
+    {
+        if (executeCommand("rpm -q presto-server-rpm") != 0) {
```
sudarsanm94: I don't think so it would be a problem.
That's why I have a wrapper to execute commands.
To implement these APIs we need to use shell commands.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
sudarsanm94: I think rebecca answered this question  on [https://github.com/jirassimok/presto-manager/pull/12](this) pr.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
sudarsanm94: The REST cal would fail if the coordinator is asked to gracefully stop. This check is needed for now I think. We can add extra checks in the controller to prevent this from happening.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
sudarsanm94: The command is indeed idempotent.
This just checks whether the command `service presto stop` executed correctly.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
jirassimok: Right.

My comment still stands for `Entity::json`, though.

And `AgentFileUtils::getFileProperty` should be given a better name so it is clear what it does.

(And `RequestBuilder::newRequestBuilder` should probably be called `create`, but that's neither here nor there.)

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
jirassimok: The problem is that the existing wrapper is very weak, and doesn't provide significant functionality.

We need shell commands, yes, and my suggestion is that we should have a more flexible system to manage them.

For example, we will certainly want to log their output, possibly to the different log files, which `executeCommand` does not provide facilities for.

(And the two minute wait...)

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
jirassimok: Then you don't need to check if it is already running here (much less throw an exception if it is).

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
jirassimok: All right.

Then this `if`-block should end the function silently (or with an "OK", if it returned a `Response`) rather than throwing an exception.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
jirassimok: Ah, that makes sense.

I recommend following the pattern from Presto installation: start the attempt and respond with "ACCEPTED", then leave it to the user to see if it actually started.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
sudarsanm94: The URI would always be `localhost:<presto_port>/v1/info/state`
So I don't think making it as a parameter is a good idea

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
sudarsanm94: Updated

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
sudarsanm94: I can do that. This would make start asynchronous. Is it fine to have one API path asynchronous and rest of the them as synchronous?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
sudarsanm94: I'll update this

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
ShanBai6: I would not let it stop if argument is wrong

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
ShanBai6: On that note, we could open a new PR for changing names once MVP is done. Now it is fine leaving it as it is.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentStatusUtils.java
ShanBai6: I feel the same, but @sudarsanm94 suggests this would be fixed for this.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
```diff
@@ -0,0 +1,220 @@
...
+            try {
+                int startPresto = executeCommand(120, "service", "presto", "startPresto");
+                if (startPresto != 0) {
+                    LOGGER.error("Failed to startPresto presto");
+                }
+            }
+            catch (PrestoManagerException e) {
+                LOGGER.error(e.getMessage(), e.getCause());
+            }
+            catch (Exception e) {
```
ShanBai6: I think there is no other exceptions to catch here

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
```diff
@@ -0,0 +1,220 @@
...
+                LOGGER.error("Unhandled exception", e);
+            }
+        }).start();
+        return Response.status(ACCEPTED).entity("Check status to make sure that presto has started").build();
```
ShanBai6: Entity should say that Presto is starting, Pls check back using status later.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    public static Response stopPresto(StopType stopType)
+    {
+        if (!isInstalled()) {
+            return Response.status(NOT_FOUND).build();
```
ShanBai6: Missing message in entity

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
```diff
@@ -0,0 +1,220 @@
...
+        }
+        if (!isRunning()) {
+            LOGGER.debug("Presto is not running");
+            return Response.status(OK).build();
```
ShanBai6: Ditto

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
```diff
@@ -0,0 +1,220 @@
...
+        }
+        try {
+            stopHelper(stopType);
+        }
+        catch (PrestoManagerException e) {
+            LOGGER.error(e.getMessage(), e.getCause());
+            return Response.status(INTERNAL_SERVER_ERROR).build();
```
ShanBai6: Ditto

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
```diff
@@ -0,0 +1,220 @@
...
+        }
+        catch (Exception e) {
+            LOGGER.error("Unhandled exception", e);
+            return Response.status(INTERNAL_SERVER_ERROR).build();
```
ShanBai6: Ditto, also may not be needing this as the method does not throw other type of exception

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
```diff
@@ -0,0 +1,220 @@
...
+                }
+                break;
+            case GRACEFUL:
+                try {
+                    if (getFileProperty(PATH_TO_CONFIG_FILE, "coordinator").equalsIgnoreCase("true")) {
+                        throw new PrestoManagerException("Coordinator can't be gracefully stopped");
+                    }
+                    String prestoPort = getFileProperty(PATH_TO_CONFIG_FILE, "http-server.http.port");
+                    Response response = newRequestBuilder(new URI(format("localhost:%s/v1/info/state", prestoPort)))
+                            .withBody(entity((new Gson().toJson("SHUTTING_DOWN")), APPLICATION_JSON_TYPE)).buildPut().invoke();
```
ShanBai6: This `Gson` object is used in multiple places, create a private field so that you could use only one. I don't feel strong about this though.

However, if the only message is a String, maybe you could just put it as Text/plain

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
```diff
@@ -0,0 +1,220 @@
...
+        }
+        catch (PrestoManagerException e) {
+            LOGGER.error(e.getMessage(), e.getCause());
+            return Response.status(INTERNAL_SERVER_ERROR).build();
+        }
+        catch (Exception e) {
+            LOGGER.error("Unhandled exception", e);
+            return Response.status(INTERNAL_SERVER_ERROR).build();
+        }
+        return Response.status(OK).build();
```
ShanBai6: Ditto

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
```diff
@@ -0,0 +1,220 @@
...
+
+    public static Response restartPresto()
+    {
+        if (!isInstalled()) {
+            return Response.status(NOT_FOUND).build();
+        }
+        try {
+            int prestoRestart = executeCommand("service", "presto", "restart");
+            if (prestoRestart != 0) {
+                LOGGER.error("Failed to restart presto; Process exited with return value: {}", prestoRestart);
```
ShanBai6: You should throw exceptions here as well, otherwise, response OK may have been sent back

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
```diff
@@ -0,0 +1,220 @@
...
+            }
+        }
+        catch (PrestoManagerException  e) {
+            LOGGER.error(e.getMessage(), e.getCause());
+            return Response.status(INTERNAL_SERVER_ERROR).build();
+        }
+        catch (Exception e) {
```
ShanBai6: Ditto

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    public static Response prestoStatus()
+    {
+        if (!isInstalled()) {
+            return Response.status(OK).entity(new Gson().toJson("Presto is not installed"))
+                    .type(APPLICATION_JSON_TYPE).build();
+        }
+        if (!isRunning()) {
+            return Response.status(OK).entity(new Gson().toJson("Presto is installed but not running"))
```
ShanBai6: You should probably provide the same info here as below although it is not running

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
```diff
@@ -0,0 +1,220 @@
...
+    public enum StopType
+    {
+        GRACEFUL, TERMINATE, KILL;
+
+        public static StopType fromString(String name)
+        {
+            try {
+                return valueOf(name.toUpperCase());
+            }
+            catch (IllegalArgumentException e) {
```
ShanBai6: I would throw the PMException here

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/ControlAPI.java
```diff
@@ -0,0 +1,102 @@
...
+
+import static com.teradata.prestomanager.agent.AgentControlUtils.prestoStatus;
+import static com.teradata.prestomanager.agent.AgentControlUtils.restartPresto;
+import static com.teradata.prestomanager.agent.AgentControlUtils.startPresto;
+import static com.teradata.prestomanager.agent.AgentControlUtils.stopPresto;
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+import static javax.ws.rs.core.MediaType.TEXT_PLAIN;
+
+@Path("/presto")
+@Api(description = "The presto API")
```
ShanBai6: Need to be more specific

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
```diff
@@ -0,0 +1,220 @@
...
+    public static Response startPresto()
+    {
+        if (!isInstalled()) {
+            return Response.status(NOT_FOUND).entity("Presto is not installed").build();
+        }
+        new Thread(() -> {
+            try {
+                int startPresto = executeCommand(120, "service", "presto", "startPresto");
+                if (startPresto != 0) {
+                    LOGGER.error("Failed to startPresto presto");
```
ShanBai6: Not english

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/ControlAPI.java
```diff
@@ -0,0 +1,102 @@
...
+@Singleton
+public class ControlAPI
```
ShanBai6: Method names could be startPresto, etc. more informative

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
```diff
@@ -0,0 +1,220 @@
...
+import static com.teradata.prestomanager.common.RequestBuilder.newRequestBuilder;
+import static java.lang.String.format;
+import static javax.ws.rs.client.Entity.entity;
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON_TYPE;
+import static javax.ws.rs.core.Response.Status.ACCEPTED;
+import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
+import static javax.ws.rs.core.Response.Status.NOT_FOUND;
+import static javax.ws.rs.core.Response.Status.OK;
+
+public final class AgentControlUtils
```
jirassimok: Maybe `PrestoControlUtils`? This sounds like it's controlling the Agent.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
sudarsanm94: Fixed

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
sudarsanm94: It catches ```IllegalArgumentException```. This makes sure that all the exceptions are caught and logged

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
sudarsanm94: Stop is not going to return any message body

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
sudarsanm94: This code is querying presto. The presto API accepts only JSON input.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
```diff
@@ -0,0 +1,220 @@
...
+{
+    private static final Logger LOGGER = LogManager.getLogger(AgentControlUtils.class);
+    private static final int WAIT_TIME_SECONDS = 60;
+    private static final int STEP_MILLISECONDS = 100;
```
jirassimok: Neither of these constants are used.

sudarsanm94: Removed.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
```diff
@@ -0,0 +1,220 @@
...
+    private static final String LAUNCHER_SCRIPT = "/usr/lib/presto/bin/launcher";
+    private static final Path PATH_TO_CONFIG_FILE = Paths.get("/etc/presto/config.properties");
```
jirassimok: `PRESTO_CONFIG_FILE` works just as well.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
sudarsanm94: I want to have ```GRACEFUL``` as default stopType

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
sudarsanm94: Changed

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
```diff
@@ -0,0 +1,220 @@
...
+            return Response.status(INTERNAL_SERVER_ERROR).entity(new Gson().toJson("Failed to get status"))
+                    .type(APPLICATION_JSON_TYPE).build();
+        }
+    }
+
+    private static boolean isInstalled()
+    {
+        try {
+            if (executeCommand("rpm", "-q", "presto-server-rpm") != 0) {
+                LOGGER.error("Presto is not installed");
```
jirassimok: This should be logged at `DEBUG` level, most likely; leave error for program errors.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
```diff
@@ -0,0 +1,220 @@
...
+            LOGGER.error("Unhandled exception", e);
+        }
+        return true;
+    }
+
+    private static boolean isRunning()
+    {
+        try {
+            if (executeCommand("service", "presto", "status") != 0) {
+                LOGGER.error("Presto is not running");
```
jirassimok: Same as in `isInstalled`.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
```diff
@@ -0,0 +1,220 @@
...
+        try {
+            if (executeCommand("rpm", "-q", "presto-server-rpm") != 0) {
+                LOGGER.error("Presto is not installed");
+                return false;
+            }
+        }
+        catch (PrestoManagerException e) {
+            LOGGER.error(e.getMessage(), e.getCause());
+        }
+        catch (Exception e) {
```
jirassimok: If there are certain exceptions that can be thrown here that you don't handle, why not let them go?

Most library functions document exactly which exceptions they can throw, so there's little risk of unexpected exceptions.

ShanBai6: I have the same suggestion

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
sudarsanm94: changed

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
sudarsanm94: Can you explain this comment?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
sudarsanm94: Fixed

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
```diff
@@ -0,0 +1,220 @@
...
+            return newRequestBuilder(new URI(format("localhost:%s/v1/info", prestoPort))).buildGet().invoke();
+        }
+        catch (IOException | URISyntaxException e) {
+            LOGGER.error("Failed to get status", e);
+            return Response.status(INTERNAL_SERVER_ERROR).entity(new Gson().toJson("Failed to get status"))
+                    .type(APPLICATION_JSON_TYPE).build();
+        }
+    }
+
+    private static boolean isInstalled()
```
jirassimok: If the command to check if Presto is installed fails, this function will indicate that Presto _is_ installed.

I would expect the behavior when a subprocess fails to be an HTTP 500 response to the API call.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
```diff
@@ -0,0 +1,220 @@
...
+        catch (PrestoManagerException e) {
+            LOGGER.error(e.getMessage(), e.getCause());
+        }
+        catch (Exception e) {
+            LOGGER.error("Unhandled exception", e);
+        }
+        return true;
+    }
+
+    private static boolean isRunning()
```
jirassimok: The same comments apply here as to `isInstalled`.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
```diff
@@ -0,0 +1,220 @@
...
+    {
+        GRACEFUL, TERMINATE, KILL;
+
+        public static StopType fromString(String name)
+        {
+            try {
+                return valueOf(name.toUpperCase());
+            }
+            catch (IllegalArgumentException e) {
+                return GRACEFUL;
```
jirassimok: Maybe make this a constant in `AgentControlUtils`, so the default can be changed more easily.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/ControlAPI.java
```diff
@@ -0,0 +1,102 @@
...
+    @Path("/start")
+    @Produces(TEXT_PLAIN)
+    @ApiOperation(value = "Start presto")
+    @ApiResponses(value = {
+            @ApiResponse(code = 202, message = "Acknowledged request"),
+            @ApiResponse(code = 404, message = "Presto is not installed")
+    })
+    public synchronized Response start()
+    {
+        LOGGER.debug("POST /presto/startPresto");
```
jirassimok: typo

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
sudarsanm94: Fixed

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
sudarsanm94: Fixed

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentControlUtils.java
ShanBai6: Should return the status/version of Presto, nvm

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoControlUtils.java
```diff
@@ -0,0 +1,182 @@
...
+    private static final String LAUNCHER_SCRIPT = "/usr/lib/presto/bin/launcher";
+    private static final Path PRESTO_CONFIG_FILE = Paths.get("/etc/presto/config.properties");
+    private static final Gson GSON = new Gson();
+
+    private PrestoControlUtils() {}
+
+    public static Response startUsingRpm()
+    {
+        try {
+            if (executeCommand("rpm", "-q", "presto-server-rpm") != 0) {
```
jirassimok: This is a matter of personal preference: you could keep the `isInstalled` function, and have it just be the condition above. 
The same goes for the Presto status command.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoControlUtils.java
```diff
@@ -0,0 +1,182 @@
...
+        return Response.status(OK).build();
+    }
+
+    public static Response statusUsingRpm()
+    {
+        try {
+            if (executeCommand("rpm", "-q", "presto-server-rpm") != 0) {
+                LOGGER.debug("Presto is not installed");
+                return Response.status(OK).entity(GSON.toJson("Presto is not installed"))
+                        .type(APPLICATION_JSON_TYPE).build();
```
jirassimok: This should probably be `NOT_FOUND`; "Presto not installed" is a status of the machine, and the API generally gives status for Presto itself.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoControlUtils.java
sudarsanm94: The ```isInstalled``` function didn't prevent the duplication of code since I removed the  ```try-catch``` block with the function. So I thought it's better to inline the function

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoControlUtils.java
sudarsanm94: For ```status```  this a correct behavior. 
Successfully retrieved the status and the status is "Presto not installed"

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoControlUtils.java
```diff
@@ -0,0 +1,182 @@
...
+            }
+        }).start();
+        return Response.status(ACCEPTED).entity("Presto is being started. Please check back later using status.").build();
+    }
+
+    public static Response stopUsingRpm(StopType stopType)
+    {
+        try {
+            if (executeCommand("rpm", "-q", "presto-server-rpm") != 0) {
+                LOGGER.debug("Presto is not installed");
```
ShanBai6: Should probably be warning. Same for all the below NOT FOUND response

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoControlUtils.java
```diff
@@ -0,0 +1,182 @@
...
+                    }
+                    String prestoPort = getFileProperty(PRESTO_CONFIG_FILE, "http-server.http.port");
+                    Response response = newRequestBuilder(new URI(format("localhost:%s/v1/info/state", prestoPort)))
+                            .withBody(entity((new Gson().toJson("SHUTTING_DOWN")), APPLICATION_JSON_TYPE)).buildPut().invoke();
+                    if (response.getStatus() != 200) {
+                        throw new PrestoManagerException("Failed to stop presto gracefully");
+                    }
+                }
+                catch (IOException | URISyntaxException e) {
+                    throw new PrestoManagerException("Error in stopping presto", e);
```
ShanBai6: Could have the same message as above. Personal preference.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoControlUtils.java
```diff
@@ -0,0 +1,197 @@
...
+
+public final class PrestoControlUtils
+{
+    private static final Logger LOGGER = LogManager.getLogger(PrestoControlUtils.class);
+    private static final String LAUNCHER_SCRIPT = "/usr/lib/presto/bin/launcher";
+    private static final Path PRESTO_CONFIG_FILE = Paths.get("/etc/presto/config.properties");
+    private static final Gson GSON = new Gson();
+    private static final int SUBPROCESS_TIMEOUT = 120;
+
+    private PrestoControlUtils() {}
```
rschlussel: maybe add the word Rpm to this class.  Then there can be a different class for the targz installation.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoControlUtils.java
```diff
@@ -0,0 +1,197 @@
...
+            stopHelper(stopType);
+        }
+        catch (PrestoManagerException e) {
+            LOGGER.error(e.getMessage(), e.getCause());
+            return Response.status(INTERNAL_SERVER_ERROR).build();
+        }
+        return Response.status(OK).build();
+    }
+
+    private static void stopHelper(StopType stopType)
```
rschlussel: use a more descriptive name.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoControlUtils.java
```diff
@@ -0,0 +1,197 @@
...
+            return Response.status(INTERNAL_SERVER_ERROR).entity(GSON.toJson("Failed to get status"))
+                    .type(APPLICATION_JSON_TYPE).build();
+        }
+    }
+
+    private static boolean isInstalled()
+            throws PrestoManagerException
+    {
+        if (executeCommand("rpm", "-q", "presto-server-rpm") != 0) {
+            LOGGER.error("Presto is not installed");
```
rschlussel: don't log here.  The method calling should decide whether this is an error and log if necessary

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoControlUtils.java
```diff
@@ -0,0 +1,197 @@
...
+            return false;
+        }
+        return true;
+    }
+
+    private static boolean isRunning()
+            throws PrestoManagerException
+    {
+        if (executeCommand("service", "presto", "status") != 0) {
+            LOGGER.debug("Presto is not running");
```
rschlussel: don't log here

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoControlUtils.java
```diff
@@ -0,0 +1,197 @@
...
+            LOGGER.error(e.getMessage(), e.getCause());
+            return Response.status(INTERNAL_SERVER_ERROR).build();
+        }
+        return Response.status(OK).build();
+    }
+
+    private static void stopHelper(StopType stopType)
+            throws PrestoManagerException
+    {
+        switch (stopType) {
```
rschlussel: it might be clearer if you put the switch statement in stopUsingRpm and had separate methods for each case.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoControlUtils.java
```diff
@@ -0,0 +1,197 @@
...
+import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
+import static javax.ws.rs.core.Response.Status.NOT_FOUND;
+import static javax.ws.rs.core.Response.Status.OK;
+
+public final class PrestoControlUtils
+{
+    private static final Logger LOGGER = LogManager.getLogger(PrestoControlUtils.class);
+    private static final String LAUNCHER_SCRIPT = "/usr/lib/presto/bin/launcher";
+    private static final Path PRESTO_CONFIG_FILE = Paths.get("/etc/presto/config.properties");
+    private static final Gson GSON = new Gson();
```
rschlussel: why is this needed?  in the package api you used text()

sudarsanm94: For some reason it doesn't work.  GSON is convenient to change text to JSON format.
I'm going to change that in the package api  in my next refactor pr.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoControlUtilsRpm.java
```diff
@@ -0,0 +1,222 @@
...
+import static javax.ws.rs.client.Entity.entity;
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+import static javax.ws.rs.core.MediaType.TEXT_PLAIN;
+import static javax.ws.rs.core.Response.Status.ACCEPTED;
+import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
+import static javax.ws.rs.core.Response.Status.NOT_FOUND;
+import static javax.ws.rs.core.Response.Status.OK;
+import static org.glassfish.jersey.client.JerseyClientBuilder.createClient;
+
+public final class PrestoControlUtilsRpm
```
rschlussel: name this PrestoRpmControlUtils.  The goal is generally for class names to be nouns, so the descriptors should be the in the same place you would put them if it were a regular english sentence.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoControlUtilsRpm.java
```diff
@@ -0,0 +1,222 @@
...
+                int startPresto = executeCommand(SUBPROCESS_TIMEOUT, "service", "presto", "start");
+                if (startPresto != 0) {
+                    LOGGER.error("Failed to start presto");
+                }
+            }
+            catch (PrestoManagerException e) {
+                LOGGER.error(e.getMessage(), e.getCause());
+            }
+        }).start();
+        return Response.status(ACCEPTED).entity("Presto is being started. Please check back later using status.").build();
```
rschlussel: Presto is being started. Please check back later using status to verify that Presto has successfully started.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoControlUtilsRpm.java
```diff
@@ -0,0 +1,222 @@
...
+                    terminatePresto();
+                    break;
+                case KILL:
+                    killPresto();
+                    break;
+                case GRACEFUL:
+                    gracefulStop();
+                    break;
+                default:
+                    LOGGER.error("Invalid stoptype: {}", stopType);
```
rschlussel: put a space between stop and type.  Or make it camel case

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoControlUtilsRpm.java
```diff
@@ -0,0 +1,222 @@
...
+
+    private static void gracefulStop()
+            throws PrestoManagerException
+    {
+        try {
+            String prestoPort = getFileProperty(PRESTO_CONFIG_FILE, "http-server.http.port");
+            UriBuilder uriBuilder = UriBuilder.fromUri(format("http://localhost:%s", prestoPort));
+            Response isCoordinator = JERSEY_CLIENT.target(uriBuilder.path("/v1/info/coordinator").build())
+                    .request(TEXT_PLAIN).buildGet().invoke();
+            if (isCoordinator.getStatus() == 200) {
```
rschlussel: this seems like it would be better off conveyed via a response status

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoControlUtilsRpm.java
rschlussel: You should return some sort of error response gode

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoRpmControlUtils.java
```diff
@@ -0,0 +1,249 @@
...
+            Response prestoState = JERSEY_CLIENT.target(uriBuilder.path("/state").build())
+                    .request(APPLICATION_JSON).buildGet().invoke();
+            JsonParser jsonParser = new JsonParser();
+            JsonObject prestoStatus = jsonParser.parse(prestoInfo.readEntity(String.class)).getAsJsonObject();
+            prestoStatus.addProperty("state", jsonParser.parse(prestoState.readEntity(String.class)).getAsString());
+            return Response.status(OK).entity(prestoStatus.toString()).type(APPLICATION_JSON).build();
+        }
+        catch (ProcessingException e) {
+            LOGGER.info("Presto is not running", e);
+            return Response.status(OK).entity(GSON.toJson("Presto is installed but not running"))
```
rschlussel: maybe return the version that's installed.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoRpmControlUtils.java
```diff
@@ -0,0 +1,249 @@
...
+                LOGGER.info("Presto is not installed");
+                return Response.status(OK).entity(GSON.toJson("Presto is not installed"))
+                        .type(APPLICATION_JSON).build();
+            }
+        }
+        catch (PrestoManagerException e) {
+            LOGGER.error(e.getMessage(), e.getCause());
+            return Response.status(INTERNAL_SERVER_ERROR).build();
+        }
+        try {
```
rschlussel: what information does this provide?  Can you put a comment at the top of the method explaining?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoRpmControlUtils.java
sudarsanm94: As we discussed before, there is no reliable way to get Presto version if Presto is not running.

rschlussel: for rpm you can query it.

sudarsanm94: Java doesn't do well with pipes. 
So querying rpm is not an option. 
There's a hack to this:
We can query the presto version `rpm -qi presto-server-rpm` ,  then save the output. Parse the output and get the presto version.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoRpmControlUtils.java
sudarsanm94: Added comment on top of the method

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoRpmControlUtils.java
rschlussel: you can get the version using the rpm command directly

`rpm -q --qf "%{VERSION}" presto-server-rpm`

or if you want a newline at the end:
`rpm -q --qf "%{VERSION}\n" presto-server-rpm`

PR #17, opened 2017-07-12
Add logger for /package API path

sudarsanm94:


presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -61,6 +64,7 @@
     public synchronized Response install(@NotNull @ApiParam("Url to fetch package") String locationToFetchPackage,
             @QueryParam("checkDependencies") @DefaultValue("true") @ApiParam("If false, disables dependency checking") boolean checkDependencies)
     {
+        LOGGER.debug("PUT /package  url: {} checkDependencies: {}", locationToFetchPackage, checkDependencies);
```
ShanBai6: For each catch here, you should log the error

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -141,7 +147,7 @@ public void run()
                 command.runCommand();
             }
             catch (Exception e) {
-                // TODO: Add to logger
+                LOGGER.error("Encountered exception", e);
```
ShanBai6: This one catch all the PrestoManagerException? If so, you should give more informative message, i.e. record the e.message in the log

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
```diff
@@ -49,20 +52,22 @@ public void runCommand()
             case RPM:
                 File tempFile;
                 try {
+                    LOGGER.debug("Downloading package from url : {}", urlToFetchPackage);
```
ShanBai6: For any debug message, one would be sufficient to tell the user its current command. If it succeed, then there is no need to say it out. Only error needs to be recorded.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -13,25 +13,41 @@
...
             throws PrestoManagerException
     {
-        try {
+        LOGGER.debug("Command executed: {}", command);
+
+        try  {
             Process process = getRuntime().exec(command);
+            Stream<String> inputStream = new BufferedReader(new InputStreamReader(process.getInputStream())).lines();
+            Stream<String> errorStream = new BufferedReader(new InputStreamReader(process.getErrorStream())).lines();
+            inputStream.forEach(line -> LOGGER.debug("{}", line));
```
ShanBai6: Is this necessary, when you call this method, you already have a line for what the command is at the API level.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
sudarsanm94: Updated

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
sudarsanm94: Yes. Whenever I throw PrestoManagerException, I add a description reg that error.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
sudarsanm94: Updated

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
sudarsanm94: Yes.
To log the shell commands executed and their output.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -88,19 +94,24 @@ public synchronized Response install(@NotNull @ApiParam("Url to fetch package")
     public synchronized Response upgrade(@NotNull @ApiParam("Url to fetch package") String locationToFetchPackage,
             @QueryParam("checkDependencies") @DefaultValue("true") @ApiParam("If false, disables dependency checking") boolean checkDependencies)
     {
+        LOGGER.debug("POST /package  url: {} checkDependencies: {}", locationToFetchPackage, checkDependencies);
         try {
-            if (installedPackageType != PackageType.valueOf(getFileExtension(locationToFetchPackage).toUpperCase())) {
+            PackageType packageType = PackageType.valueOf(getFileExtension(locationToFetchPackage).toUpperCase());
+            if (installedPackageType != packageType) {
+                LOGGER.error("Expected package type {} , found {}", installedPackageType, packageType);
```
ShanBai6: The message should be like: `Unmatched package type` or something like `Installed using blablabla, but upgrade using blablabla`

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
ShanBai6: Will this record the error message along with the exception in the log?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -61,17 +64,20 @@
     public synchronized Response install(@NotNull @ApiParam("Url to fetch package") String locationToFetchPackage,
             @QueryParam("checkDependencies") @DefaultValue("true") @ApiParam("If false, disables dependency checking") boolean checkDependencies)
     {
+        LOGGER.debug("PUT /package  url: {} checkDependencies: {}", locationToFetchPackage, checkDependencies);
         try {
             installedPackageType = PackageType.valueOf(getFileExtension(locationToFetchPackage).toUpperCase());
         }
         catch (IllegalArgumentException e) {
+            LOGGER.error("Invalid package type for presto-server. Package must be .tar.gz or .rpm", e);
```
ShanBai6: For all package type mismatch, you should log what is the type expected & what is given

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -13,25 +13,41 @@
...
 {
+    private static final Logger LOGGER = LogManager.getLogger(CommandExecutor.class);
+
     private CommandExecutor() {}
 
     public static int executeCommand(String command)
             throws PrestoManagerException
     {
-        try {
+        LOGGER.debug("Command executed: {}", command);
```
jirassimok: This should be "Command about to execute" or something similar.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
```diff
@@ -81,9 +84,10 @@ private static void upgradeUsingRpm(String pathToRpm, boolean disableDependencyC
         if (disableDependencyChecking) {
             nodeps = "--nodeps";
         }
-        String upgradeRpm = format("sudo rpm -U %s %s", nodeps, pathToRpm);
+        String upgradeRpm = format("sudo rpm -Uv %s %s", nodeps, pathToRpm);
```
ShanBai6: I am not sure if you should use -v mode. As you are feeding everything in the inputstream to log. Keep it precise should be the goal

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
jirassimok: I strongly recommend against this type of logging; the subprocess' output should not have separate log entries for each line, it should be logged all together, and probably in a different file from most of the Presto Manager logs.

And it should be logged at the `info` level, I think.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -13,25 +13,41 @@
...
+
+        try  {
             Process process = getRuntime().exec(command);
+            Stream<String> inputStream = new BufferedReader(new InputStreamReader(process.getInputStream())).lines();
+            Stream<String> errorStream = new BufferedReader(new InputStreamReader(process.getErrorStream())).lines();
+            inputStream.forEach(line -> LOGGER.debug("{}", line));
+            errorStream.forEach(line -> LOGGER.debug("{}", line));
```
jirassimok: These should be logged at the `error` or `warning` level. I'm not sure which.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
sudarsanm94: Having -v flag provides useful information like
Package already installed. 
It provides two lines of useful information for debugging.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
sudarsanm94: The error message will be recorded in the  exception variable e

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -13,25 +13,41 @@
...
             process.waitFor(2, TimeUnit.MINUTES);
-            return process.exitValue();
+            int retValue = process.exitValue();
+            LOGGER.debug("Process exited with return value {}", retValue);
```
jirassimok: I would put a colon every time a value is introduced in the log, in case the value is empty or otherwise unclear.

(This is less relevant for this particular line, but quite relevant at other lines.)

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
```diff
@@ -81,9 +84,10 @@ private static void upgradeUsingRpm(String pathToRpm, boolean disableDependencyC
         if (disableDependencyChecking) {
             nodeps = "--nodeps";
         }
```
jirassimok: The previous four lines could be simplified with a ternary operator.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -61,17 +64,20 @@
     public synchronized Response install(@NotNull @ApiParam("Url to fetch package") String locationToFetchPackage,
             @QueryParam("checkDependencies") @DefaultValue("true") @ApiParam("If false, disables dependency checking") boolean checkDependencies)
     {
+        LOGGER.debug("PUT /package  url: {} checkDependencies: {}", locationToFetchPackage, checkDependencies);
```
jirassimok: There should be some indicator of the end of each parameter. I recommend putting a semicolon or comma after each `{}`.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
jirassimok: I would include the URL here, so the invalid package can be diagnosed.

"Invalid package '[URL]'; expected '.tar' or '.rpm'."

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -61,17 +64,20 @@
...
         try {
             installedPackageType = PackageType.valueOf(getFileExtension(locationToFetchPackage).toUpperCase());
```
jirassimok: This isn't relevant yet, but using `valueOf` will not work for tarballs.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
jirassimok: I would add an additional `catch` before this one, catching `PrestoManagerException` specifically, and logging with `error(e.getMessage(), e.getCause())`. Then the message from the exception is logged, and the exception it wraps (if there is one) will be logged (if there isn't one, nothing happens, at least when I tested it).

Then the current `catch` could log `error("Unhandled exception", e)`.

ShanBai6: You should fix this

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
jirassimok: @sudarsanm94 

I think that we should hold off on adding logging here until we know how we should log subprocess output.

(And see my previous comment.)

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -13,25 +13,41 @@
...
     {
-        try {
+        LOGGER.debug("Command to be executed: {}", command);
+
+        try  {
             Process process = getRuntime().exec(command);
+            Stream<String> inputStream = new BufferedReader(new InputStreamReader(process.getInputStream())).lines();
+            Stream<String> errorStream = new BufferedReader(new InputStreamReader(process.getErrorStream())).lines();
+            inputStream.forEach(line -> LOGGER.info("{}", line));
+            errorStream.forEach(line -> LOGGER.warn("{}", line));
```
jirassimok: I think that we should hold off on adding logging here until we know how we should log subprocess output.

And my previous comment on these two lines, which is still relevant:

> I strongly recommend against this type of logging; the subprocess' output should not have separate log entries for each line, it should be logged all together, and probably in a different file from most of the Presto Manager logs.

jirassimok:
My comment in the second thread still stands.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -61,17 +65,20 @@
     public synchronized Response install(@NotNull @ApiParam("Url to fetch package") String locationToFetchPackage,
```
jirassimok: This method should be annotated with `@Consumes({MediaType.TEXT_PLAIN})`.

The same goes the other methods in this class.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
jirassimok: `Stream.collect(Collectors.joining(System.getProperty("line.separator"))` will concatenate an entire output stream into a single string that you can log.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUninstaller.java
```diff
@@ -40,7 +43,7 @@ public void runCommand()
             case RPM:
                 uninstallUsingRpm("presto-server-rpm", disableDependencyChecking, ignoreNotInstalled);
                 break;
-            case TARBALL:
+            case GZ:
```
jirassimok: This should be `TARBALL` for now.

We can fix how that is handled later, when we add support for tarballs.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -13,25 +13,48 @@
...
+            String output = new BufferedReader(new InputStreamReader(process.getInputStream()))
+                    .lines().collect(Collectors.joining(System.getProperty("line.separator")));
+            String error = new BufferedReader(new InputStreamReader(process.getErrorStream()))
+                    .lines().collect(Collectors.joining(System.getProperty("line.separator")));
+            LOGGER.info("Standard output from command : {} \n{}", command, output);
+            if ("".equals(error)) {
+                LOGGER.info("No standard error from command : {}", command);
+            }
+            else {
+                LOGGER.error("Standard error from command : {} \n{}", command, error);
```
rschlussel: also log when you're done waiting for the process

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -13,25 +13,48 @@
...
+            }
             process.waitFor(2, TimeUnit.MINUTES);
-            return process.exitValue();
+            int retValue = process.exitValue();
+            LOGGER.debug("Process exited with return value: {}", retValue);
```
rschlussel: probably want more than debug info if the process returned with a non-zero error code.  Maybe throw a presto manager exception including the error code and the stderr output.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -115,13 +128,14 @@ public synchronized Response uninstall(
...
+        LOGGER.debug("DELETE /package checkDependencies: {} ; ignoreErrors: {}", checkDependencies, ignoreErrors);
         new Thread(new PrestoAsynchronousCommand(new PrestoUninstaller(installedPackageType, checkDependencies, ignoreErrors))).start();
-        return Response.status(ACCEPTED).entity(text("Presto is being uninstalled")).build();
+        return Response.status(ACCEPTED).entity("Presto is being uninstalled").build();
     }
 
     public enum PackageType
     {
-        RPM, TARBALL;
+        RPM, GZ
```
rschlussel: unrelated put in refactoring commit

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
```diff
@@ -49,35 +52,37 @@ public void runCommand()
...
-                // TODO: Add to logger
                 throw new IllegalArgumentException(format("Unsupported package type %s", packageType));
         }
     }
 
     private static void installUsingRpm(String pathToRpm, boolean checkDependencies)
             throws PrestoManagerException
     {
-        String checkRpm = format("rpm -K --nosignature %s", pathToRpm);
+        String checkRpm = format("rpm -Kv --nosignature %s", pathToRpm);
```
sudarsanm94: So that the output of the command would be looged

rschlussel: oh you mean when you capture standard out.  I see.  I was confused because the logging wasn't happening in this method.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
sudarsanm94: If the process returned with a non-zero error code then the method which calls commandexecutor will throw PrestoMangerException

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
sudarsanm94: We can know that the process has finished by this log
```LOGGER.debug("Process exited with return value: {}", retValue);```

File: presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java (hunk 18 above)
Comment (rschlussel) [inline presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java]: then that's where the logging should happen too.

Comment (sudarsanm94) [inline presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java]: The PrestoMangerException would be caught eventually and the error would be logged

File: presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java [hunk 21]
```diff
@@ -13,23 +13,42 @@
...
     private CommandExecutor() {}
 
     public static int executeCommand(String command)
             throws PrestoManagerException
     {
+        LOGGER.debug("Command to be executed: {}", command);
+
         try {
             Process process = getRuntime().exec(command);
+            String output = new BufferedReader(new InputStreamReader(process.getInputStream()))
```
rschlussel: 1. You don't close these
2. These will hang around until the process exits, which isn't good (And also makes the waitFor kind of useless since it's already waited).  Also, it will first do the stdout and then the stderr, which causes its own problems.  It's probably worth redirecting stderr to stdout so you don't need to worry about reading from both of them (They'll both get logged anyway, keeping them separate isn't so important).

You should use processBuilder to help you out here.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -13,23 +13,42 @@
...
+                    .lines().collect(Collectors.joining(System.getProperty("line.separator")));
+            String error = new BufferedReader(new InputStreamReader(process.getErrorStream()))
+                    .lines().collect(Collectors.joining(System.getProperty("line.separator")));
+            LOGGER.info("Standard output from command : {} \n{}", command, output);
+            if (!"".equals(error)) {
+                LOGGER.warn("Standard error from command : {} \n{}", command, error);
+            }
             process.waitFor(2, TimeUnit.MINUTES);
```
rschlussel: I don't think this closes the process if it hasn't exited, it just waits around for that amount of time or until the process exits, whichever comes first (haven't tried it out, just based on the docs).  You may want to force terminate the process.  in any case process.exitValue() will throw an error if the process hasn't exited.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -13,26 +13,53 @@
...
+        StringTokenizer tokenizer = new StringTokenizer(command);
+        String[] commandArray = new String[tokenizer.countTokens()];
+        for (int i = 0; tokenizer.hasMoreTokens(); i++) {
+            commandArray[i] = tokenizer.nextToken();
+        }
+
+        ProcessBuilder processBuilder = new ProcessBuilder(commandArray);
+        processBuilder.redirectErrorStream(true);
+        processBuilder.redirectOutput(appendTo(logFile));
+        Process process = null;
```
rschlussel: You don't need to explicitly set new Object variables to null.  That's the default for objects if you don't set them to anything.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -13,26 +13,53 @@
...
         try {
-            Process process = getRuntime().exec(command);
+            process = processBuilder.start();
             process.waitFor(2, TimeUnit.MINUTES);
             return process.exitValue();
         }
+        catch (IllegalThreadStateException e) {
+            throw new PrestoManagerException(format("Command %s timed out", command));
+        }
         catch (Exception e) {
```
rschlussel: What exceptions could happen here?  Explicitly list any expected exceptions and handle those.  Don't just catch everything.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -13,26 +13,53 @@
...
+            commandArray[i] = tokenizer.nextToken();
+        }
+
+        ProcessBuilder processBuilder = new ProcessBuilder(commandArray);
+        processBuilder.redirectErrorStream(true);
+        processBuilder.redirectOutput(appendTo(logFile));
+        Process process = null;
         try {
-            Process process = getRuntime().exec(command);
+            process = processBuilder.start();
```
rschlussel: instead of the try/catch, do:
```
process = processBuilder.start()
if(!process.waitFor(2, TimeUnit.MINUTES)) {
     process.destroyForcibly();
}
return process.exitValue();
```

sudarsanm94: How do you get the exitvalue if process is destroyed?

rschlussel: when you kill a process it still returns an exit code

rschlussel: though we may want to throw the timeout exception there instead

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -13,25 +13,55 @@
...
+        ProcessBuilder processBuilder = new ProcessBuilder(commandArray);
+        processBuilder.redirectErrorStream(true);
         try {
-            Process process = getRuntime().exec(command);
-            process.waitFor(2, TimeUnit.MINUTES);
+            Process process = processBuilder.start();
+            try (InputStream processStream = process.getInputStream()) {
+                new Thread(() -> {
+                    try {
+                        LOGGER.info("Output from command : {} \n {}", command, new String(IOUtils.toByteArray(processStream)));
```
jirassimok: Use `com.google.common.io.ByteStreams#toByteArray(InputStream)` rather than Apache.

sudarsanm94: Is there any reason to use one over the other?

jirassimok: We already have Guava as a dependency.

sudarsanm94: We already have common.io as a dependency.

rschlussel: guava's more standard.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -13,25 +13,55 @@
...
+                    }
+                    catch (IOException e) {
+                        LOGGER.error("Failed to log the process output");
+                    }
+                }).start();
+                if (!process.waitFor(2, TimeUnit.MINUTES)) {
+                    process.destroyForcibly();
```
jirassimok: I would try using `destroy()` first. Then wait for another short time, check if it's alive, and destroy it forcibly if it is (and log that it resisted the first attempt).

With the builder idea, the delay between the two calls should be settable.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -13,25 +13,55 @@
...
     public static int executeCommand(String command)
             throws PrestoManagerException
     {
+        LOGGER.debug("Command to be executed: {}", command);
+
+        StringTokenizer tokenizer = new StringTokenizer(command);
+        String[] commandArray = new String[tokenizer.countTokens()];
+        for (int i = 0; tokenizer.hasMoreTokens(); i++) {
+            commandArray[i] = tokenizer.nextToken();
+        }
```
jirassimok: `String#split(String)` is probably a better choice than a `StringTokenizer`.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
jirassimok: And we use Guava more; we might be able to remove `commons.io`; if I recall correctly, it's only used in one other place.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
sudarsanm94: This is the way getRuntime.exec() does it. So I thought this would be better

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -13,25 +13,55 @@
...
+                    try {
+                        LOGGER.info("Output from command : {} \n {}", command, new String(IOUtils.toByteArray(processStream)));
+                    }
+                    catch (IOException e) {
+                        LOGGER.error("Failed to log the process output");
+                    }
+                }).start();
+                if (!process.waitFor(2, TimeUnit.MINUTES)) {
+                    process.destroyForcibly();
+                    throw new PrestoManagerException(format("Command %s timed out", command));
```
jirassimok: Put some type of quotes around the command, or put it at the end after a colon.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
jirassimok: There shouldn't be a space before the colon.
And there shouldn't be a space on either side of the newline.
Similar advice applies to various other log messages in this pull request.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
sudarsanm94: From the java docs 
```destroyForcibly()``` invokes ```destroy()```. So I think its not much of a difference

File: presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java [hunk 30]
```diff
@@ -62,22 +66,22 @@ public void runCommand()
...
-            throw new PrestoManagerException("Corrupted RPM");
+        int checkRpm = executeCommand(format("rpm -Kv --nosignature %s", pathToRpm));
+        if (checkRpm != 0) {
+            throw new PrestoManagerException("Corrupted RPM", checkRpm);
         }
         String nodeps = checkDependencies ? "" : "--nodeps";
-        String installRpm = format("sudo rpm -i %s %s", nodeps, pathToRpm);
-        if (executeCommand(installRpm) != 0) {
-            throw new PrestoManagerException("Failed to install presto");
+        int installRpm = executeCommand(format("sudo rpm -iv %s %s", nodeps, pathToRpm));
```
jirassimok: `pathToRpm` may cause problems if it has an unusual name.

The most important reason to pass the arguments to a subprocess as separate strings is because of this sort of thing.

Imagine if the path provided was `--prefix=/sbin init.rpm`.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -13,44 +13,79 @@
...
+    private int timeout;
 
-    private CommandExecutor() {}
+    private CommandExecutor(String[] commandArray)
+    {
+        this.commandArray = requireNonNull(commandArray);
+        timeout = 60;
+    }
+
+    public static CommandExecutor newCommandExecutor(String tokenizeCommand, String... args)
```
jirassimok: I would just call this `commandExecutor`, or just `command`, but that's a matter of taste.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
jirassimok: Not necessarily.

On UNIX, `destroy` eventually leads to SIGTERM, while `destroyForcibly` leads to SIGKILL. If the process has significant resources, the difference could matter.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -13,44 +13,79 @@
...
 {
     private static final Logger LOGGER = LogManager.getLogger(CommandExecutor.class);
+    private final String[] commandArray;
+    private int timeout;
 
-    private CommandExecutor() {}
+    private CommandExecutor(String[] commandArray)
+    {
+        this.commandArray = requireNonNull(commandArray);
+        timeout = 60;
```
jirassimok: This should still be configurable by the caller.
```
public CommandExecutor setTimeout(int seconds)
{
    this.timeout = seconds;
    return this;
}
```

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
sudarsanm94: From Java docs of destroyForcibly:
The default implementation of this method invokes destroy() and so may not forcibly terminate the process.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -13,25 +13,79 @@
...
+    private int timeout;
+
+    private CommandExecutor(String[] commandArray)
+    {
+        this.commandArray = requireNonNull(commandArray);
+        timeout = 60;
+    }
 
-    public static int executeCommand(String command)
+    public static CommandExecutor commandExecutor(String tokenizeCommand, String... args)
```
rschlussel: This isn't a very intuitive breakdown, and I think it isn't needed.   Instead make the other constructor public and pass everything in already tokenized

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -13,25 +13,79 @@
...
+            commandList.add(tokenizer.nextToken());
+        }
+        for (String s : args) {
+            commandList.add(s);
+        }
+        String[] commandArray = commandList.toArray(new String[commandList.size()]);
+        return new CommandExecutor(commandArray);
+    }
+
+    public CommandExecutor setTimeout(int timeoutInSeconds)
```
rschlussel: make the timeout final and return a new command executor with the timeout set  (so create a private constructor that takes a tokenized command and a timeout that this and the other constructor call).   And call this withTimeout() instead of setTimeout().

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -13,26 +13,80 @@
...
+        return new CommandExecutor(toCommandArray(command), 60).execute();
+    }
+
+    static int commandExecutor(int timeoutInSeconds, String... command)
+            throws PrestoManagerException
+    {
+        return new CommandExecutor(toCommandArray(command), timeoutInSeconds).execute();
+    }
+
+    private static String[] toCommandArray(String... command)
```
jirassimok: What is this function's purpose? Does it do anything besides turn an array into a list and back again?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -13,26 +13,80 @@
...
+        this.timeout = timeout;
+    }
+
+    static int commandExecutor(String... command)
             throws PrestoManagerException
     {
+        return new CommandExecutor(toCommandArray(command), 60).execute();
+    }
+
+    static int commandExecutor(int timeoutInSeconds, String... command)
```
jirassimok: This should be `executeCommand`.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -13,26 +13,80 @@
...
 {
-    private CommandExecutor() {}
+    private static final Logger LOGGER = LogManager.getLogger(CommandExecutor.class);
+    private final String[] commandArray;
+    private final int timeout;
 
-    public static int executeCommand(String command)
+    private CommandExecutor(String[] commandArray, int timeout)
+    {
+        this.commandArray = requireNonNull(commandArray);
```
jirassimok: This should probably be something like `requireNonEmpty` (except that doesn't exist).

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -13,26 +13,80 @@
...
+        this.timeout = timeout;
+    }
+
+    static int commandExecutor(String... command)
             throws PrestoManagerException
     {
+        return new CommandExecutor(toCommandArray(command), 60).execute();
```
jirassimok: Make default timeout a `static final` class constant, for later configuration.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -13,26 +13,73 @@
...
+    private static final Logger LOGGER = LogManager.getLogger(CommandExecutor.class);
+    private static final int defaultTimeout = 60;
+    private final String[] commandArray;
+    private final int timeout;
+
+    private CommandExecutor(String[] commandArray, int timeout)
+            throws PrestoManagerException
+    {
+        if (commandArray.length == 0) {
+            throw new PrestoManagerException("Command array is empty");
```
jirassimok: This should most likely be an `IllegalArgumentException`.
With both of my two new comments fixed, I approve this PR.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -13,26 +13,73 @@
...
 
-import static java.lang.Runtime.getRuntime;
 import static java.lang.String.format;
 
-// TODO: Add logger
 public final class CommandExecutor
 {
-    private CommandExecutor() {}
+    private static final Logger LOGGER = LogManager.getLogger(CommandExecutor.class);
+    private static final int defaultTimeout = 60;
```
jirassimok: Make this `DEFAULT_TIMEOUT`.

ShanBai6:
There could be another improvement on this, make the overload function executeCommand to have Optional parameter, that could save one of the two method.

PR #19, opened 2017-07-18
Fix log API bugs and add API call to list log files

jirassimok:
Second commit: Add a logger.

### DIFF
--- a/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
+++ b/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
@@ -20,7 +20,6 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayDeque;
-import java.util.ConcurrentModificationException;
 import java.util.Deque;
 import java.util.Map;
 import java.util.StringJoiner;
@@ -43,39 +42,51 @@ public class LogFilter
 {
     private final Path file;
     private final Pattern logPattern;
+    private final String defaultEntry;
     private final String lineSeparator;
     private final ImmutableMap<String, Predicate<String>> namedGroupFilters;
-    private final ImmutableMap<Integer, Predicate<String>> numberedGroupFilters;
     private final int maxEntries;
     private final boolean keepFirst;
 
-    private LogFilter(Path file, Pattern logPattern, String lineSeparator,
+    private LogFilter(Path file, Pattern logPattern,
+            String defaultEntry, String lineSeparator,
             Map<String, Predicate<String>> namedGroupFilters,
-            Map<Integer, Predicate<String>> numberedGroupFilters,
             int maxEntries, boolean keepFirst)
             throws FileNotFoundException
     {
         this.file = requireNonNull(file);
         this.logPattern = requireNonNull(logPattern);
+        this.defaultEntry = requireNonNull(defaultEntry);
         this.lineSeparator = requireNonNull(lineSeparator);
         this.namedGroupFilters = ImmutableMap.copyOf(requireNonNull(namedGroupFilters));
-        this.numberedGroupFilters = ImmutableMap.copyOf(requireNonNull(numberedGroupFilters));
         this.maxEntries = maxEntries;
         this.keepFirst = keepFirst;
-        if (!Files.exists(file)) {
+        if (!Files.isRegularFile(file)) {
             throw new FileNotFoundException(file.toString());
+        }
+
+        Matcher matcher = logPattern.matcher(defaultEntry);
+        if (!matcher.matches()) {
+            throw new IllegalArgumentException(
+                    "Default entry does not match log entry pattern");
+        }
+        try {
+            for (Map.Entry<String, Predicate<String>> f
+                    : namedGroupFilters.entrySet()) {
+                matcher.group(f.getKey());
+            }
+        }
+        catch (IllegalArgumentException e) {
+            throw new IllegalArgumentException(
+                    "Invalid capturing group in log filter", e);
         }
     }
 
     public Stream<String> streamEntries()
-            throws IOException, ConcurrentModificationException
+            throws IOException
     {
-        LogEntries logEntryBuilders = Files.lines(file).collect(
-                LogEntries::new,
-                LogEntries::addLine,
-                (a, b) -> {
-                    throw new ConcurrentModificationException();
-                });
+        LogEntries logEntryBuilders = new LogEntries();
+        Files.lines(file).forEachOrdered(logEntryBuilders::addLine);
         return logEntryBuilders.streamEntries().map(StringJoiner::toString);
     }
 
@@ -94,20 +105,23 @@ public class LogFilter
     {
         private Path file;
         private Pattern logPattern;
+        private String defaultEntry;
         private String lineSeparator = System.getProperty("line.separator");
-        private ImmutableMap.Builder<String, Predicate<String>> namedGroupFilters = ImmutableMap.builder();
-        private ImmutableMap.Builder<Integer, Predicate<String>> numberedGroupFilters = ImmutableMap.builder();
+        private ImmutableMap.Builder<String, Predicate<String>> namedGroupFilters;
         private int maxEntries = Integer.MAX_VALUE;
-        private boolean keepFirst = false;
+        private boolean keepFirst;
 
-        private Builder() {}
+        private Builder()
+        {
+            namedGroupFilters = ImmutableMap.builder();
+        }
 
         public LogFilter build()
                 throws FileNotFoundException
         {
-            return new LogFilter(file, logPattern, lineSeparator,
+            return new LogFilter(file, logPattern,
+                    defaultEntry, lineSeparator,
                     namedGroupFilters.build(),
-                    numberedGroupFilters.build(),
                     maxEntries, keepFirst);
         }
 
@@ -127,25 +141,28 @@ public class LogFilter
         }
 
         /**
-         * Add a filter to apply to a named capturing group
-         * in the regular expression for log file lines
+         * Set an entry that will be inserted if the first entry does
+         * not match the entry pattern.
+         *
+         */
+        public Builder setDefaultEntry(String entry)
+        {
+            defaultEntry = requireNonNull(entry);
+            return this;
+        }
+
+        /**
+         * Add a filter to apply to a named capturing group in the
+         * regular expression for log file lines.
+         *
+         * A filter of `s -> true' will not have the same effect
+         * as having no filter; only if there is no filter will
+         * a mis-formatted initial log entry be included in results.
          */
         public Builder addGroupFilter(String groupName, Predicate<String> filter)
         {
             namedGroupFilters.put(
                     requireNonNull(groupName),
-                    requireNonNull(filter));
-            return this;
-        }
-
-        /**
-         * Add a filter to apply to a capturing group in
-         * the regular expression for log file lines
-         */
-        public Builder addGroupFilter(Integer groupNumber, Predicate<String> filter)
-        {
-            numberedGroupFilters.put(
-                    requireNonNull(groupNumber),
                     requireNonNull(filter));
             return this;
         }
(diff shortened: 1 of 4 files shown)


presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
```diff
@@ -43,28 +43,53 @@
...
             throw new FileNotFoundException(file.toString());
         }
+
+        Matcher matcher = logPattern.matcher(defaultEntry);
+        if (!matcher.matches()) {
+            throw new IllegalArgumentException(
+                    "Default entry does not match log entry pattern");
+        }
+        try {
+            for (Map.Entry<String, Predicate<String>> f
```
ShanBai6: You have a checkFilter method for this whole thing

ShanBai6: Also you should add error handling to that method as well.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
```diff
@@ -75,62 +105,72 @@ public static Response getLogs(String filename, JaxrsParameter<Instant> startDat
...
         }
 
         Stream<String> logEntries;
         try {
             logEntries = logFilter.streamEntries();
         }
         catch (IOException e) {
-            return Response.status(Status.INTERNAL_SERVER_ERROR)
-                    .entity("IOException while reading file").build();
+            return internalWarning("IOException while reading file", e);
```
ShanBai6: You did not catch the `ConcurrentModificationException` thrown by streamEntries() here, ditto for another same method call few lines below.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
```diff
@@ -245,4 +293,28 @@ else if (nullEnd) {
             return s -> level.name().equals(s);
         }
     }
+
```
ShanBai6: You could put those in a separate utility class and make it support more functions so that we could standardize it.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
```diff
@@ -56,13 +63,36 @@
...
+    private static final String DEFAULT_ENTRY = "0000-01-01T00:00:00.000+0000" +
+            "\tALL\t[none]\t[none]\tThis log entry was not preceded by a header:";
 
     private LogsHandler() {}
 
+    public static Response getLogList()
+    {
+        String fileList;
+        try (Stream<java.nio.file.Path> files = Files.list(LOG_DIRECTORY)) {
+            fileList = files.filter(Files::isRegularFile)
```
jirassimok: I generally try to avoid making assumptions like that, in case there are log files with unusual names.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
jirassimok: I hadn't caught it because the exception should never occur, but it can't hurt to be careful; changed.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
jirassimok: `checkFilters` short-circuits; I need to make sure all of the filters are legal with the default entry, `checkFilters` would only tell me if they all pass.

There's no error handling in `checkFilters` because by the time it is called, this block has already run, ensuring that the filters' names and indices are all valid.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
```diff
@@ -139,8 +185,9 @@ public Builder addGroupFilter(String groupName, Predicate<String> filter)
         }
 
         /**
-         * Add a filter to apply to a capturing group in
-         * the regular expression for log file lines
+         * Identical to {@link #addGroupFilter(String, Predicate)},
```
jirassimok: And numbered groups won't do much good when you're matching against a user-specified regular expression. Removed.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
jirassimok: If you mean the various `internalError` and similar functions, I'm not sure that's a great idea; combining the logging and responses should be done carefully to make sure it is appropriate for the situation at hand.

Though I think I _will_ make a utility class for common responses.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
ShanBai6: Basically we decide how log files are named, so should not worry about that

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
ShanBai6: Then why do you throw it. Delete that may also be plausible.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
ShanBai6: Or look into libraries, maybe there is one ready for use. Just like the `HtttpMethod` enum.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
jirassimok: `stream().collect(suppler, accumulator, combiner)` requires a "combiner" for parallel streams, but I am using a sequential stream, and the operation I am performing can not be performed in parallel.

But your comments have made me reconsider using `collect` here at all, and I've changed the implementation to something a bit nicer.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
jirassimok: Well, there's a few methods on `Response` that make it easier to create responses `ok`, `accepted`, and `serverError` to name a few.

I've made a utility class with methods for three of the most common errors. I left out methods for the success codes, because those are used less frequently (about once per API function), and the main purpose of the class is to make code look nicer by saving space.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
jirassimok: We don't get to name Presto's log files.
There shouldn't be any harm in this; the log directory doesn't tent to contain files that it shouldn't.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
```diff
@@ -75,67 +109,75 @@ public static Response getLogs(String filename, JaxrsParameter<Instant> startDat
...
-                    .entity(Files.exists(filePath)
-                            ? "Not a regular file"
-                            : "File not found")
-                    .build();
+        Path filePath;
+        try {
+            filePath = LOG_DIRECTORY.resolve(filename);
+        }
+        catch (InvalidPathException e) {
+            return badRequest("Invalid file name");
```
ShanBai6: Could also be NotFound, since it is about file

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
```diff
@@ -245,4 +293,17 @@ else if (nullEnd) {
             return s -> level.name().equals(s);
         }
     }
+
+    private static Response internalError(
```
ShanBai6: You could migrate these two methods to the utility as well. Do the logging in the utility is also acceptable

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
jirassimok: This is for when the path itself is invalid, like if there was a slash in it.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
jirassimok: That would cause the log entries to list the utility class as their source, which wouldn't be very useful. Though that could be solved in a few ways...

I'll look into better solutions later today, if I have time.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/LogsAPI.java
```diff
@@ -58,8 +59,17 @@
             .toFormatter().withChronology(IsoChronology.INSTANCE);
 
     @GET
+    @Produces({MediaType.TEXT_PLAIN})
+    @ApiOperation(value = "Get a listing of log files")
+    @ApiResponses({@ApiResponse(code = 200, message = "Retrieved file list")})
+    public Response getLogList()
+    {
+        return LogsHandler.getLogList();
```
sudarsanm94: Add logger  ```GET logs/```

jirassimok: Done

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
```diff
@@ -43,7 +46,7 @@
 public final class LogsHandler
 {
     // TODO: Make _all_ of these 'private static final' values configurable
-    private static final Path LOG_DIRECTORY = Paths.get("/tmp", "presto", "logs");
+    private static final Path LOG_DIRECTORY = Paths.get("test", "presto", "logs");
```
cawallin: why test?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogFilter.java
```diff
@@ -20,7 +20,6 @@
 import java.nio.file.Files;
```
cawallin: Ideally, you should divide up this commit ("Fix logs API...") to have one commit with fixing the logs API and one with adding the logs list. (It seems like there are a couple of other assorted bug fixes, mostly around whether paths exist -- those should be either in a separate commit or at least mentioned in the commit message)

You don't need to change this now, though, just keep it in mind going into the future.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
cawallin: Give the stacktrace too, so that whoever's debugging it knows how to fix it

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
```diff
@@ -107,28 +107,23 @@ public static Response getLogs(String filename, JaxrsParameter<Instant> startDat
...
-                        .entity("Can not provide date range and limit number of entries")
-                        .build();
+                return badRequest(
+                        "Can not provide date range and limit number of entries");
             }
             else if (start.isAfter(end)) {
-                return Response.status(Status.BAD_REQUEST)
-                        .entity("End of date range can not be before start")
-                        .build();
+                return badRequest("End of date range can not be before start");
```
cawallin: include start and end date in the error message

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
```diff
@@ -107,28 +107,23 @@ public static Response getLogs(String filename, JaxrsParameter<Instant> startDat
         requireNonNull(logLevel);
 
         if (!startDate.isValid() || !endDate.isValid()) {
-            return Response.status(Status.BAD_REQUEST)
-                    .entity("Invalid date format").build();
+            return badRequest("Invalid date format");
```
cawallin: maybe include the date format that was given in the error message

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
jirassimok: Send it back, or log it?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
jirassimok: At this point, the expected log format has been lost.

I could add a "get error message" method to `JaxrsParameter`, though, and include relevant information there.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
jirassimok: Done.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
```diff
@@ -164,9 +215,8 @@ public static Response deleteLogs(String filename, JaxrsParameter<Instant> endDa
                 return Response.status(Status.NO_CONTENT).build();
             }
             catch (IOException e) {
-                // TODO: Log exception, log file disappeared since check for 404
                 return Response.status(Status.INTERNAL_SERVER_ERROR)
-                        .entity("File disappeared while processing operation").build();
+                        .entity("IOException while writing file").build();
```
jirassimok: It's in the log

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
jirassimok: It's a convenient placeholder name.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
jirassimok: I found a better solution; I can throw a `WebApplicationException` when parsing the parameters, with the desired response as the argument. I'll do that.

jirassimok: This will go in a future pull request.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
cawallin: Send it back (maybe truncated) and log the full thing -- much easier for the user

jirassimok: I've added a future TODO to add exception messages to responses throughout the application.

PR #20, opened 2017-07-20
Refactor presto install, uninstall and upgrade methods

sudarsanm94:
1. Remove package type checking
2. Remove ```commons.io``` dependency
3. Check whether presto is running before install and upgrade

### DIFF
--- a/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
+++ b/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
@@ -13,36 +13,33 @@
  */
 package com.teradata.prestomanager.agent;
 
-import com.teradata.prestomanager.agent.api.PackageAPI.PackageType;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 import java.io.File;
-import java.io.IOException;
 import java.net.URL;
 
 import static com.teradata.prestomanager.agent.CommandExecutor.executeCommand;
-import static java.io.File.createTempFile;
+import static com.teradata.prestomanager.agent.PackageApiUtils.checkRpmPackage;
+import static com.teradata.prestomanager.agent.PackageApiUtils.fetchFileFromUrl;
 import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
-import static org.apache.commons.io.FileUtils.copyURLToFile;
 
 public final class PrestoInstaller
         implements PrestoCommand
 {
     private static final Logger LOGGER = LogManager.getLogger(PrestoInstaller.class);
-    private static final int CONNECTION_TIMEOUT = 1000000;
-    private static final int READ_TIMEOUT = 10000;
+    private static final int SUBPROCESS_TIMEOUT = 150;
 
     private final PackageType packageType;
     private final URL urlToFetchPackage;
-    private final boolean disableDependencyChecking;
+    private final boolean checkDependencies;
 
-    public PrestoInstaller(PackageType packageType, URL urlToFetchPackage, boolean disableDependencyChecking)
+    public PrestoInstaller(PackageType packageType, URL urlToFetchPackage, boolean checkDependencies)
     {
         this.packageType = requireNonNull(packageType);
         this.urlToFetchPackage = requireNonNull(urlToFetchPackage);
-        this.disableDependencyChecking = requireNonNull(disableDependencyChecking);
+        this.checkDependencies = requireNonNull(checkDependencies);
     }
 
     public void runCommand()
@@ -50,17 +47,15 @@ public final class PrestoInstaller
     {
         switch (packageType) {
             case RPM:
-                File tempFile;
+                File tempFile = fetchFileFromUrl(urlToFetchPackage);
                 try {
-                    LOGGER.debug("Downloading package from url: {}", urlToFetchPackage);
-                    tempFile = createTempFile("presto", ".rpm");
-                    copyURLToFile(urlToFetchPackage, tempFile, CONNECTION_TIMEOUT, READ_TIMEOUT);
+                    installUsingRpm(tempFile.toString(), checkDependencies);
                 }
-                catch (IOException e) {
-                    throw new PrestoManagerException(format("Failed to download file from url %s", urlToFetchPackage), e);
+                finally {
+                    if (!tempFile.delete()) {
+                        LOGGER.warn("Failed to delete the tempFile: {}", tempFile.toString());
+                    }
                 }
-                installUsingRpm(tempFile.toString(), disableDependencyChecking);
-                tempFile.delete();
                 break;
             case TARBALL:
                 // TODO: Add tarball installation
@@ -73,12 +68,14 @@ public final class PrestoInstaller
     private static void installUsingRpm(String pathToRpm, boolean checkDependencies)
             throws PrestoManagerException
     {
-        int checkRpm = executeCommand("rpm", "-Kv", "--nosignature ", pathToRpm);
-        if (checkRpm != 0) {
-            throw new PrestoManagerException("Corrupted RPM", checkRpm);
+        checkRpmPackage(pathToRpm);
+        int installRpm;
+        if (checkDependencies) {
+            installRpm = executeCommand(SUBPROCESS_TIMEOUT, "sudo", "rpm", "-iv", pathToRpm);
         }
-        String nodeps = checkDependencies ? "" : "--nodeps";
-        int installRpm = executeCommand(150, "sudo", "rpm", "-iv", nodeps, pathToRpm);
+        else {
+            installRpm = executeCommand(SUBPROCESS_TIMEOUT, "sudo", "rpm", "-iv", "--nodeps", pathToRpm);
+        }
         if (installRpm != 0) {
             throw new PrestoManagerException("Failed to install presto", installRpm);
         }
--- a/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
+++ b/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
@@ -17,18 +17,25 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 import java.io.File;
+import java.io.IOException;
 import java.net.URL;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 
 import static com.teradata.prestomanager.agent.CommandExecutor.executeCommand;
 import static com.teradata.prestomanager.agent.PackageApiUtils.checkRpmPackage;
 import static com.teradata.prestomanager.agent.PackageApiUtils.fetchFileFromUrl;
 import static java.lang.String.format;
+import static java.nio.file.Files.createDirectory;
+import static java.nio.file.Files.isDirectory;
+import static java.nio.file.Files.write;
 import static java.util.Objects.requireNonNull;
 
 public final class PrestoInstaller
         implements PrestoCommand
 {
     private static final Logger LOGGER = LogManager.getLogger(PrestoInstaller.class);
+    private static final Path CATALOG_DIR = Paths.get("/etc/presto/catalog");
     private static final int SUBPROCESS_TIMEOUT = 150;
 
     private final PackageType packageType;
@@ -79,6 +86,21 @@ public final class PrestoInstaller
         if (installRpm != 0) {
             throw new PrestoManagerException("Failed to install presto", installRpm);
         }
+        addTpchCatalog();
         LOGGER.debug("Successfully installed presto");
+    }
+
+    private static void addTpchCatalog()
+            throws PrestoManagerException
+    {
+        try {
+            if (!isDirectory(CATALOG_DIR)) {
+                createDirectory(CATALOG_DIR);
+                write(CATALOG_DIR.resolve("tpch.properties"), "connector.name=tpch".getBytes());
+            }
+        }
+        catch (IOException e) {
+            throw new PrestoManagerException("Failed to add tpch catalog", e);
+        }
     }
 }
--- a/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUninstaller.java
+++ b/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUninstaller.java
@@ -13,7 +13,6 @@
  */
 package com.teradata.prestomanager.agent;
 
-import com.teradata.prestomanager.agent.api.PackageAPI.PackageType;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
@@ -25,6 +24,9 @@ public class PrestoUninstaller
         implements PrestoCommand
 {
     private static final Logger LOGGER = LogManager.getLogger(PrestoUninstaller.class);
+    private static final String PRESTO_PACKAGE = "presto-server-rpm";
+    private static final int SUBPROCESS_TIMEOUT = 90;
+
     private final PackageType packageType;
     private final boolean disableDependencyChecking;
     private final boolean ignoreNotInstalled;
@@ -41,25 +43,25 @@ public class PrestoUninstaller
     {
         switch (packageType) {
             case RPM:
-                uninstallUsingRpm("presto-server-rpm", disableDependencyChecking, ignoreNotInstalled);
+                uninstallUsingRpm(disableDependencyChecking, ignoreNotInstalled);
                 break;
             case TARBALL:
-                // TODO: Add tarball installation
+                // TODO: Add tarball uninstallation
                 throw new UnsupportedOperationException("Tarball uninstall is not supported");
             default:
                 throw new IllegalArgumentException(format("Unsupported package type %s", packageType));
         }
     }
 
-    private static void uninstallUsingRpm(String packageName, boolean checkDependencies, boolean ignoreErrors)
+    private static void uninstallUsingRpm(boolean checkDependencies, boolean ignoreErrors)
             throws PrestoManagerException
     {
-        int checkPackageInstalled = executeCommand("sudo", "rpm", "-q", packageName);
+        int checkPackageInstalled = executeCommand("sudo", "rpm", "-q", PRESTO_PACKAGE);
         if (checkPackageInstalled != 0) {
             if (!ignoreErrors) {
-                throw new PrestoManagerException(format("Package %s is not installed", packageName), checkPackageInstalled);
+                throw new PrestoManagerException(format("Package '%s' is not installed", PRESTO_PACKAGE), checkPackageInstalled);
             }
-            LOGGER.warn("Package {} is not installed; Process exited with return value: {}", packageName, checkPackageInstalled);
+            LOGGER.warn("Package '{}' is not installed; Process exited with return value: {}", PRESTO_PACKAGE, checkPackageInstalled);
             return;
         }
         if (executeCommand("service", "presto", "status") == 0) {
@@ -68,12 +70,15 @@ public class PrestoUninstaller
             }
             LOGGER.warn("Presto is running");
         }
+        int uninstallPackage;
+        if (checkDependencies) {
+            uninstallPackage = executeCommand(SUBPROCESS_TIMEOUT, "sudo", "rpm", "-e", PRESTO_PACKAGE);
+        }
         else {
-            String nodeps = checkDependencies ? "" : "--nodeps";
-            int uninstallPackage = executeCommand(90, "sudo", "rpm", "-e", nodeps, packageName);
-            if (uninstallPackage != 0) {
-                throw new PrestoManagerException(format("Failed to uninstall package %s", packageName), uninstallPackage);
-            }
+            uninstallPackage = executeCommand(SUBPROCESS_TIMEOUT, "sudo", "rpm", "-e", "--nodeps", PRESTO_PACKAGE);
+        }
+        if (uninstallPackage != 0) {
+            throw new PrestoManagerException(format("Failed to uninstall package: %s", PRESTO_PACKAGE), uninstallPackage);
         }
         LOGGER.debug("Successfully uninstalled presto");
     }
--- a/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
+++ b/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
@@ -13,36 +13,33 @@
  */
 package com.teradata.prestomanager.agent;
 
-import com.teradata.prestomanager.agent.api.PackageAPI.PackageType;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 import java.io.File;
-import java.io.IOException;
 import java.net.URL;
 
 import static com.teradata.prestomanager.agent.CommandExecutor.executeCommand;
-import static java.io.File.createTempFile;
+import static com.teradata.prestomanager.agent.PackageApiUtils.checkRpmPackage;
+import static com.teradata.prestomanager.agent.PackageApiUtils.fetchFileFromUrl;
 import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
-import static org.apache.commons.io.FileUtils.copyURLToFile;
 
 public class PrestoUpgrader
         implements PrestoCommand
 {
     private static final Logger LOGGER = LogManager.getLogger(PrestoUpgrader.class);
-    private static final int CONNECTION_TIMEOUT = 1000000;
-    private static final int READ_TIMEOUT = 10000;
+    private static final int SUBPROCESS_TIMEOUT = 150;
 
     private final PackageType packageType;
     private final URL urlToFetchPackage;
-    private final boolean disableDependencyChecking;
+    private final boolean checkDependencies;
 
-    public PrestoUpgrader(PackageType packageType, URL urlToFetchPackage, boolean disableDependencyChecking)
+    public PrestoUpgrader(PackageType packageType, URL urlToFetchPackage, boolean checkDependencies)
     {
         this.packageType = requireNonNull(packageType);
         this.urlToFetchPackage = requireNonNull(urlToFetchPackage);
-        this.disableDependencyChecking = requireNonNull(disableDependencyChecking);
+        this.checkDependencies = requireNonNull(checkDependencies);
     }
 
     public void runCommand()
@@ -50,20 +47,15 @@ public class PrestoUpgrader
     {
         switch (packageType) {
             case RPM:
-                if (executeCommand("service", "presto", "status") == 0) {
-                    throw new PrestoManagerException("Presto is running");
-                }
-                File tempFile;
+                File tempFile = fetchFileFromUrl(urlToFetchPackage);
                 try {
-                    LOGGER.debug("Downloading package from url: {}", urlToFetchPackage);
-                    tempFile = createTempFile("presto", ".rpm");
-                    copyURLToFile(urlToFetchPackage, tempFile, CONNECTION_TIMEOUT, READ_TIMEOUT);
+                    upgradeUsingRpm(tempFile.toString(), checkDependencies);
                 }
-                catch (IOException e) {
-                    throw new PrestoManagerException(format("Failed to download file from url %s", urlToFetchPackage), e);
+                finally {
+                    if (!tempFile.delete()) {
+                        LOGGER.warn("Failed to delete the tempFile: {}", tempFile.toString());
+                    }
                 }
-                upgradeUsingRpm(tempFile.toString(), disableDependencyChecking);
-                tempFile.delete();
                 break;
             case TARBALL:
                 // TODO: Add tarball installation
@@ -73,18 +65,17 @@ public class PrestoUpgrader
         }
     }
 
-    private static void upgradeUsingRpm(String pathToRpm, boolean disableDependencyChecking)
+    private static void upgradeUsingRpm(String pathToRpm, boolean checkDependencies)
             throws PrestoManagerException
     {
-        int checkRpm = executeCommand("rpm", "-Kv", "--nosignature", pathToRpm);
-        if (checkRpm != 0) {
-            throw new PrestoManagerException("Corrupted RPM", checkRpm);
+        checkRpmPackage(pathToRpm);
+        int upgradeRpm;
+        if (checkDependencies) {
+            upgradeRpm = executeCommand(SUBPROCESS_TIMEOUT, "sudo", "rpm", "-U", pathToRpm);
         }
-        String nodeps = "";
-        if (disableDependencyChecking) {
-            nodeps = "--nodeps";
+        else {
+            upgradeRpm = executeCommand(SUBPROCESS_TIMEOUT, "sudo", "rpm", "-U", "--nodeps", pathToRpm);
         }
-        int upgradeRpm = executeCommand(150, "sudo", "rpm", "-U", nodeps, pathToRpm);
         if (upgradeRpm != 0) {
             throw new PrestoManagerException("Failed to upgrade presto", upgradeRpm);
         }
--- a/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
+++ b/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
@@ -17,11 +17,15 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 import java.io.File;
+import java.io.IOException;
 import java.net.URL;
+import java.nio.file.Paths;
 
+import static com.google.common.io.MoreFiles.deleteDirectoryContents;
 import static com.teradata.prestomanager.agent.CommandExecutor.executeCommand;
 import static com.teradata.prestomanager.agent.PackageApiUtils.checkRpmPackage;
 import static com.teradata.prestomanager.agent.PackageApiUtils.fetchFileFromUrl;
+import static java.io.File.createTempFile;
 import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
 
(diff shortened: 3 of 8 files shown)


presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -36,87 +36,86 @@
...
+    @Consumes({TEXT_PLAIN})
+    @Produces({TEXT_PLAIN})
     @ApiOperation(value = "Install presto using rpm or tarball")
     @ApiResponses(value = {
             @ApiResponse(code = 202, message = "Acknowledged request"),
-            @ApiResponse(code = 400, message = "Bad package format or Invalid url")
+            @ApiResponse(code = 409, message = "Presto is running"),
+            @ApiResponse(code = 400, message = "Invalid url")
     })
     public synchronized Response install(@NotNull @ApiParam("Url to fetch package") String locationToFetchPackage,
```
jirassimok: Using `@NotNull` here doesn't really do anything.

The methods of this class shouldn't be called by hand, so it's not particularly useful as documentation (and if we want it for that, we should have many more such annotations). And Jersey doesn't care about the `NotNull` annotation; if it passes in `null`, we need to check anyway.

Speaking of which, I believe that it will use an empty string, not `null`, when given no body, and we could check for that. "Expected URL in body" is more useful as a response than "Invalid URL".

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
```diff
@@ -73,15 +68,38 @@ public void runCommand()
...
         String nodeps = checkDependencies ? "" : "--nodeps";
-        int installRpm = executeCommand(150, "sudo", "rpm", "-iv", nodeps, pathToRpm);
+        int installRpm = executeCommand(SUBPROCESS_TIMEOUT, "sudo", "rpm", "-iv", nodeps, pathToRpm);
         if (installRpm != 0) {
             throw new PrestoManagerException("Failed to install presto", installRpm);
         }
         LOGGER.debug("Successfully installed presto");
     }
+
+    static File fetchFileFromUrl(URL urlToFetchPackage)
```
jirassimok: This is very, very similar to a method in `AgentFileUtils`. Unfortunately, it is different enough that that method can't be used here, but we should make note of that for a future update, where we can make `AgentFileUtils` a more robust class and use it more widely.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUninstaller.java
```diff
@@ -70,9 +73,9 @@ private static void uninstallUsingRpm(String packageName, boolean checkDependenc
         }
```
ShanBai6: You should add return here. Because the success message should not be logged when Presto is running

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -36,87 +36,86 @@
...
-    @Consumes({MediaType.TEXT_PLAIN})
-    @Produces({MediaType.TEXT_PLAIN})
+    @Consumes({TEXT_PLAIN})
+    @Produces({TEXT_PLAIN})
     @ApiOperation(value = "Install presto using rpm or tarball")
     @ApiResponses(value = {
             @ApiResponse(code = 202, message = "Acknowledged request"),
-            @ApiResponse(code = 400, message = "Bad package format or Invalid url")
+            @ApiResponse(code = 409, message = "Presto is running"),
+            @ApiResponse(code = 400, message = "Invalid url")
```
ShanBai6: Order by response code number

ShanBai6: Ditto below

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -129,13 +128,13 @@ public synchronized Response uninstall(
...
-        new Thread(new PrestoAsynchronousCommand(new PrestoUninstaller(installedPackageType, checkDependencies, ignoreErrors))).start();
-        return Response.status(ACCEPTED).entity(text("Presto is being uninstalled")).build();
+        new Thread(new PrestoAsynchronousCommand(new PrestoUninstaller(PackageType.RPM, checkDependencies, ignoreErrors))).start();
+        return Response.status(ACCEPTED).entity("Presto is being uninstalled").build();
     }
 
     public enum PackageType
     {
-        RPM, TARBALL;
+        RPM, TARBALL
```
ShanBai6: You probably want this enum somewhere else, should be in some utility/ a class of its own. i.e enum class.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -162,4 +161,16 @@ public void run()
             }
         }
     }
+
+    private static boolean isRunning()
+            throws PrestoManagerException
+    {
+        if (executeCommand("rpm", "-q", "presto-server-rpm") == 0) {
+            if (executeCommand("service", "presto", "status") == 0) {
+                LOGGER.error("Presto is running");
```
ShanBai6: Should be warning

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
sudarsanm94: I think this enum is not big enough to have it as a separate class. If we add more functionalities to this class in the future then we can make it as a separate class

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
sudarsanm94: I choose ```error``` because we send be ```conflict``` message

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
sudarsanm94: Changed

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
sudarsanm94: Changed

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
```diff
@@ -74,6 +77,7 @@ private static void installUsingRpm(String pathToRpm, boolean checkDependencies)
         if (installRpm != 0) {
             throw new PrestoManagerException("Failed to install presto", installRpm);
         }
+        addTpchCatalog();
```
jirassimok: What if the user doesn't want to have any catalog files?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
```diff
@@ -67,10 +72,42 @@ private static void upgradeUsingRpm(String pathToRpm, boolean checkDependencies)
...
+            throws PrestoManagerException
+    {
+        Path tempConfigTar;
+        try {
+            tempConfigTar = Files.createTempFile("presto_config", ".tar.gz");
+        }
+        catch (IOException e) {
+            throw new PrestoManagerException("Failed to create a temp tar file", e);
+        }
+        int tarConfigDir = executeCommand("tar", "-czv", "-f", tempConfigTar.toString(), "-C", CONFIG_DIR.toString(), ".");
```
jirassimok: There's no reason to `tar` the files here.

And you aren't using your temporary file; you're creating a new file with the same name in your current directory.

`tar -czv -f tempFile configDir` would be better, if you need to use `tar`.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
```diff
@@ -67,10 +72,42 @@ private static void upgradeUsingRpm(String pathToRpm, boolean checkDependencies)
...
+        if (tarConfigDir != 0) {
+            throw new PrestoManagerException("Failed to gather the contents of config directory", tarConfigDir);
+        }
+        return tempConfigTar.toString();
+    }
+
+    private static void deployConfigDir(String tempConfigTar)
+            throws PrestoManagerException
+    {
+        int deleteConfigDir = executeCommand("rm", "-rf", CONFIG_DIR.resolve("*").toString());
```
jirassimok: This deletes a file named `*`.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
sudarsanm94: The user can delete the catalog if it is not required. But I think tpch is a basic catalog which is needed.

jirassimok: Testing isn't the standard use case.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUninstaller.java
```diff
@@ -68,12 +71,15 @@ private static void uninstallUsingRpm(String packageName, boolean checkDependenc
             }
             LOGGER.warn("Presto is running");
         }
+        int uninstallPackage;
+        if (checkDependencies) {
+            uninstallPackage = executeCommand(SUBPROCESS_TIMEOUT, "sudo", "rpm", "-e", PRESTO_PACKAGE);
+        }
         else {
-            String nodeps = checkDependencies ? "" : "--nodeps";
-            int uninstallPackage = executeCommand(90, "sudo", "rpm", "-e", nodeps, packageName);
```
jirassimok: Have you tested with `checkDependencies` false?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -36,87 +35,94 @@
...
+            LOGGER.error("Failed to ascertain whether presto is running", e);
+            return Response.status(INTERNAL_SERVER_ERROR).entity("Failed to ascertain whether presto is running").build();
         }
         catch (MalformedURLException e) {
             LOGGER.error("Invalid url: {}", locationToFetchPackage, e);
-            return Response.status(BAD_REQUEST).entity(text("Invalid url")).build();
+            return Response.status(BAD_REQUEST).entity("Invalid url").build();
         }
-        return Response.status(ACCEPTED).entity(text("Check status to make sure that presto is upgraded")).build();
+        return Response.status(ACCEPTED).entity("Presto is being upgraded. Please check back later using status.").build();
```
jirassimok: Either don't specify how to check back, or be more explicit.
`"Presto is being upgraded. Check installation status at '../presto/status'"`

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUninstaller.java
sudarsanm94: Yes

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
sudarsanm94: It is convenient for the user.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -36,87 +36,96 @@
...
-        }
-        catch (IllegalArgumentException e) {
-            LOGGER.error("Invalid package: {}. Package must be .tar.gz or .rpm", locationToFetchPackage, e);
-            return Response.status(BAD_REQUEST).entity(text("Invalid package type for presto-server. Package must be .tar.gz or .rpm")).build();
+        if ("".equals(locationToFetchPackage)) {
+            LOGGER.error("Url is empty or null");
+            return Response.status(BAD_REQUEST).entity("Expected URL in the body").build();
         }
         try {
+            if (isRunning()) {
```
sudarsanm94: Its used to handle the case when user sends install command and the presto is already running.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -36,87 +36,96 @@
...
+            LOGGER.error("Failed to ascertain whether presto is running", e);
+            return Response.status(INTERNAL_SERVER_ERROR).entity("Failed to ascertain whether presto is running").build();
         }
         catch (MalformedURLException e) {
             LOGGER.error("Invalid url: {}", locationToFetchPackage, e);
-            return Response.status(BAD_REQUEST).entity(text("Invalid url")).build();
+            return Response.status(BAD_REQUEST).entity("Invalid url").build();
         }
-        return Response.status(ACCEPTED).entity(text("Check status to make sure that presto is upgraded")).build();
+        return Response.status(ACCEPTED).entity("Presto is being upgraded. Please check back later using status api.").build();
```
sudarsanm94: I don't know any other way to check whether presto is upgraded. The only way to check this is to check the version. But there is no reliable way to check the version without starting presto.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
```diff
@@ -69,16 +74,57 @@ private static void upgradeUsingRpm(String pathToRpm, boolean checkDependencies)
...
         }
         LOGGER.debug("Successfully upgraded presto");
     }
+
+    private static File tarConfigFiles()
+            throws PrestoManagerException
+    {
+        File tempFile;
+        try {
+            tempFile = File.createTempFile("ConfigFiles", ".tar.gz");
```
ShanBai6: Should include Presto here to distinguish with PM

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
```diff
@@ -69,16 +74,57 @@ private static void upgradeUsingRpm(String pathToRpm, boolean checkDependencies)
             throws PrestoManagerException
     {
         checkRpmPackage(pathToRpm);
-        int upgradeRpm;
-        if (checkDependencies) {
-            upgradeRpm = executeCommand(SUBPROCESS_TIMEOUT, "sudo", "rpm", "-U", pathToRpm);
+        File configFiles = tarConfigFiles();
```
ShanBai6: Change method name. i.e Preserve/Save

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
```diff
@@ -30,6 +34,7 @@
 {
     private static final Logger LOGGER = LogManager.getLogger(PrestoUpgrader.class);
     private static final int SUBPROCESS_TIMEOUT = 150;
+    private static final Path CONFIG_DIR = Paths.get("/etc/presto");
```
ShanBai6: Just use String if you always use it as String

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
```diff
@@ -69,16 +74,57 @@ private static void upgradeUsingRpm(String pathToRpm, boolean checkDependencies)
             throws PrestoManagerException
     {
         checkRpmPackage(pathToRpm);
-        int upgradeRpm;
-        if (checkDependencies) {
-            upgradeRpm = executeCommand(SUBPROCESS_TIMEOUT, "sudo", "rpm", "-U", pathToRpm);
+        File configFiles = tarConfigFiles();
+        try {
+            int upgradeRpm;
+            if (checkDependencies) {
```
ShanBai6: This preserve config files should be a flag like --nodeps. Let user to choose as he/she may want a fresh copy to start over

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
ShanBai6: Make sure this file is created in a directory that won't affect the upgrade process

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
```diff
@@ -69,16 +74,57 @@ private static void upgradeUsingRpm(String pathToRpm, boolean checkDependencies)
...
+    {
+        File tempFile;
+        try {
+            tempFile = File.createTempFile("ConfigFiles", ".tar.gz");
+        }
+        catch (IOException e) {
+            throw new PrestoManagerException("Failed to create temp file.", e);
+        }
+        int tarConfigFiles = executeCommand("tar", "-czv", "-f", tempFile.getAbsolutePath(), "-C", CONFIG_DIR.toString());
+        if (tarConfigFiles != 0) {
```
ShanBai6: You could inline these 2.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
```diff
@@ -69,16 +74,57 @@ private static void upgradeUsingRpm(String pathToRpm, boolean checkDependencies)
...
+            throws PrestoManagerException
+    {
+        try {
+            deleteDirectoryContents(CONFIG_DIR);
+        }
+        catch (IOException e) {
+            throw new PrestoManagerException(format("Failed to delete contents of the directory: %s", CONFIG_DIR.toString()), e);
+        }
+        int untarConfigFiles = executeCommand("tar", "-xzv", "-f", configFiles.getAbsolutePath(), "-C", CONFIG_DIR.toString());
+        if (untarConfigFiles != 0) {
```
ShanBai6: ditto

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
```diff
@@ -69,16 +74,57 @@ private static void upgradeUsingRpm(String pathToRpm, boolean checkDependencies)
...
+            if (checkDependencies) {
+                upgradeRpm = executeCommand(SUBPROCESS_TIMEOUT, "sudo", "rpm", "-U", pathToRpm);
+            }
+            else {
+                upgradeRpm = executeCommand(SUBPROCESS_TIMEOUT, "sudo", "rpm", "-U", "--nodeps", pathToRpm);
+            }
+            if (upgradeRpm != 0) {
+                throw new PrestoManagerException("Failed to upgrade presto", upgradeRpm);
+            }
+            deployConfigFiles(configFiles);
```
ShanBai6: This is probably reemploy instead of deploy

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
sudarsanm94: The file would be created in `/tmp` folder

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
```diff
@@ -69,16 +74,57 @@ private static void upgradeUsingRpm(String pathToRpm, boolean checkDependencies)
...
         }
-        else {
-            upgradeRpm = executeCommand(SUBPROCESS_TIMEOUT, "sudo", "rpm", "-U", "--nodeps", pathToRpm);
-        }
-        if (upgradeRpm != 0) {
-            throw new PrestoManagerException("Failed to upgrade presto", upgradeRpm);
+        finally {
+            if (!configFiles.delete()) {
+                LOGGER.warn("Failed to delete the tempFile: {}", configFiles.toString());
```
jirassimok: Write out "temporary file" fully.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
```diff
@@ -69,16 +74,57 @@ private static void upgradeUsingRpm(String pathToRpm, boolean checkDependencies)
...
+            throws PrestoManagerException
+    {
+        File tempFile;
+        try {
+            tempFile = File.createTempFile("ConfigFiles", ".tar.gz");
+        }
+        catch (IOException e) {
+            throw new PrestoManagerException("Failed to create temp file.", e);
+        }
+        int tarConfigFiles = executeCommand("tar", "-czv", "-f", tempFile.getAbsolutePath(), "-C", CONFIG_DIR.toString());
```
jirassimok: `-czv` and `-f` can be combined into `-czvf`. Put the `"."` back at the end or it won't work.

I would still recommend `cp -r CONFIG_DIR tempDir`, though.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
```diff
@@ -79,6 +106,37 @@ private static void upgradeUsingRpm(String pathToRpm, boolean checkDependencies)
...
+            throw new PrestoManagerException("Failed to create temp file.", e);
+        }
+        int tarConfigFiles = executeCommand("tar", "-czv", "-f", tempFile.getAbsolutePath(), "-C", CONFIG_DIR, ".");
+        if (tarConfigFiles != 0) {
+            throw new PrestoManagerException("Failed to tar config files.", tarConfigFiles);
+        }
+        return tempFile;
+    }
+
+    private static void deployConfigFiles(File configFiles)
```
jirassimok: I'd call this `configTar` or something similarly indicative of the file's expected format.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -104,7 +104,8 @@ public synchronized Response install(@ApiParam("Url to fetch package") String lo
             @ApiResponse(code = 409, message = "Presto is running")
     })
     public synchronized Response upgrade(@ApiParam("Url to fetch package") String locationToFetchPackage,
-            @QueryParam("checkDependencies") @DefaultValue("true") @ApiParam("If false, disables dependency checking") boolean checkDependencies)
+            @QueryParam("checkDependencies") @DefaultValue("true") @ApiParam("If false, disables dependency checking") boolean checkDependencies,
+            @QueryParam("preserveConfigs") @DefaultValue("false") @ApiParam("If true, preserves all the config files") boolean preserveConfigs)
```
jirassimok: Maybe call this "keepConfig"? It's a bit more succinct, and has the same meaning.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
```diff
@@ -30,16 +33,19 @@
...
     private static final int SUBPROCESS_TIMEOUT = 150;
+    private static final String CONFIG_DIR = "/etc/presto";
 
     private final PackageType packageType;
     private final URL urlToFetchPackage;
     private final boolean disableDependencyChecking;
+    private final boolean preserveConfigs;
 
-    public PrestoUpgrader(PackageType packageType, URL urlToFetchPackage, boolean disableDependencyChecking)
+    public PrestoUpgrader(PackageType packageType, URL urlToFetchPackage, boolean disableDependencyChecking, boolean preserveConfigs)
```
jirassimok: This is starting to get well into the realm of the [boolean trap](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap).
But if you'd rather not change them, please add Javadoc comments documenting the any opaque boolean parameters.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
```diff
@@ -79,6 +106,37 @@ private static void upgradeUsingRpm(String pathToRpm, boolean checkDependencies)
...
+            throws PrestoManagerException
+    {
+        File tempFile;
+        try {
+            tempFile = File.createTempFile("PrestoConfigs", ".tar.gz");
+        }
+        catch (IOException e) {
+            throw new PrestoManagerException("Failed to create temp file.", e);
+        }
+        int tarConfigFiles = executeCommand("tar", "-czv", "-f", tempFile.getAbsolutePath(), "-C", CONFIG_DIR, ".");
```
jirassimok: Maybe call this `tarResult`? (Same comment for other `executeCommand` return values.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
jirassimok: I might not; that requires one of two bad-looking changes—either a condition with the operator very far to the right, or with the constant on the left-hand side.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
```diff
@@ -79,6 +106,37 @@ private static void upgradeUsingRpm(String pathToRpm, boolean checkDependencies)
...
+    private static void deployConfigFiles(File configFiles)
+            throws PrestoManagerException
+    {
+        try {
+            deleteDirectoryContents(Paths.get(CONFIG_DIR));
+        }
+        catch (IOException e) {
+            throw new PrestoManagerException(format("Failed to delete contents of the directory: %s", CONFIG_DIR), e);
+        }
+        int untarConfigFiles = executeCommand("tar", "-xzv", "-f", configFiles.getAbsolutePath(), "-C", CONFIG_DIR);
```
jirassimok: Put the dot back here; it doesn't work like this.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -36,87 +35,100 @@
...
             @ApiResponse(code = 202, message = "Acknowledged request"),
-            @ApiResponse(code = 400, message = "Bad package format or Invalid url")
+            @ApiResponse(code = 400, message = "Invalid url"),
+            @ApiResponse(code = 409, message = "Presto is running")
     })
-    public synchronized Response upgrade(@NotNull @ApiParam("Url to fetch package") String locationToFetchPackage,
-            @QueryParam("checkDependencies") @DefaultValue("true") @ApiParam("If false, disables dependency checking") boolean checkDependencies)
+    public synchronized Response upgrade(@ApiParam("Url to fetch package") String locationToFetchPackage,
+            @QueryParam("checkDependencies") @DefaultValue("true") @ApiParam("If false, disables dependency checking") boolean checkDependencies,
+            @QueryParam("retainConfigs") @DefaultValue("true") @ApiParam("If false, config files are not retained") boolean retainConfigs)
```
jirassimok: I would say "preserve" was better than "retain" ("save" or "keep" would also work well). This only matters for the value in the `QueryParam` annotation.
Actually, reversing the boolean might be a better idea: "resetConfig" is about as clear and the others, and defaults to `false`, which is the default value for `boolean`s, which makes it easier for the user to guess the default, and more importantly, should save the user from typos (assuming that it uses `Boolean#parseBoolean(String)` to convert the parameter to a string).

And please remove the "s", we're only dealing with one configuration.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
```diff
@@ -13,57 +13,56 @@
...
+    private static final int SUBPROCESS_TIMEOUT = 150;
+    private static final String CONFIG_DIR = "/etc/presto";
 
     private final PackageType packageType;
     private final URL urlToFetchPackage;
     private final boolean disableDependencyChecking;
+    private final boolean retainConfigs;
 
-    public PrestoUpgrader(PackageType packageType, URL urlToFetchPackage, boolean disableDependencyChecking)
+    public PrestoUpgrader(PackageType packageType, URL urlToFetchPackage, boolean disableDependencyChecking, boolean retainConfigs)
```
jirassimok: I suspect that it would be a better design decision to not pass the package type here, and instead use a different class to manage tarball installations.

Also, `urlToFetchPackage` could be shortened to `packageUrl`. (And I would reverse `disableDependencyChecking` to `checkDependencies` for both brevity and to avoid the double negative.)

```java
/*
 * @param packageType The type of package to install.
 * @param packageUrl The URL from which to retrieve Presto.
 * @param checkDependencies Whether to check RPM dependencies before an RPM upgrade.
 * @param retainConfig Whether to preserve the existing configuration files.
 */
```

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
```diff
@@ -73,21 +72,72 @@ public void runCommand()
...
+    private static void deployConfigFiles(File configTar)
+            throws PrestoManagerException
+    {
+        try {
+            deleteDirectoryContents(Paths.get(CONFIG_DIR));
+        }
+        catch (IOException e) {
+            throw new PrestoManagerException(format("Failed to delete contents of the directory: %s", CONFIG_DIR), e);
+        }
+        int untarResult = executeCommand("tar", "-xzvf", configTar.getAbsolutePath(), "-C", CONFIG_DIR);
```
jirassimok: This one still needs the dot back.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -36,87 +35,100 @@
...
+            if (isInstalled()) {
+                LOGGER.error("Presto is already installed.");
+                return Response.status(CONFLICT).entity("Presto is already installed.").build();
+            }
             URL url = new URL(locationToFetchPackage);
-            new Thread(new PrestoAsynchronousCommand(new PrestoInstaller(installedPackageType, url, checkDependencies))).start();
+            new Thread(new PrestoAsynchronousCommand(new PrestoInstaller(RPM, url, checkDependencies))).start();
+        }
+        catch (PrestoManagerException e) {
+            LOGGER.error("Failed to ascertain whether presto is running", e);
```
rschlussel: should say installed, not running

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -36,87 +35,100 @@
...
+        if ("".equals(locationToFetchPackage)) {
+            LOGGER.error("Url is empty or null");
+            return Response.status(BAD_REQUEST).entity("Expected URL in the body").build();
+        }
         try {
-            PackageType packageType = PackageType.valueOf(getFileExtension(locationToFetchPackage).toUpperCase());
-            if (installedPackageType != packageType) {
-                LOGGER.error("Package type mismatch. Presto installed using {} , but attempting to upgrade using {}", installedPackageType, packageType);
-                return Response.status(BAD_REQUEST).entity(text("Provided package format doesn't match the installed package format")).build();
+            if (isRunning()) {
```
rschlussel: You shouldn't need to stop presto first.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -36,87 +35,100 @@
...
+            return Response.status(INTERNAL_SERVER_ERROR).entity("Failed to ascertain whether presto is already installed.").build();
         }
         catch (MalformedURLException e) {
             LOGGER.error("Invalid url: {}", locationToFetchPackage, e);
-            return Response.status(BAD_REQUEST).entity(text("Invalid url")).build();
+            return Response.status(BAD_REQUEST).entity("Invalid url").build();
         }
-        return Response.status(ACCEPTED).entity(text("Check status to make sure that presto is installed")).build();
+        return Response.status(ACCEPTED).entity("Presto is being installed. " +
+                "Please check back later using status api to make sure that install is successful.").build();
```
rschlussel: *the* status api

Also, can that be verified using status?  If not remove that bit of the message.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -129,13 +141,8 @@ public synchronized Response uninstall(
...
         LOGGER.debug("DELETE /package checkDependencies: {} ; ignoreErrors: {}", checkDependencies, ignoreErrors);
-        new Thread(new PrestoAsynchronousCommand(new PrestoUninstaller(installedPackageType, checkDependencies, ignoreErrors))).start();
-        return Response.status(ACCEPTED).entity(text("Presto is being uninstalled")).build();
-    }
-
-    public enum PackageType
-    {
-        RPM, TARBALL;
+        new Thread(new PrestoAsynchronousCommand(new PrestoUninstaller(RPM, checkDependencies, ignoreErrors))).start();
+        return Response.status(ACCEPTED).entity("Presto is being uninstalled. Please check back later using status api.").build();
```
rschlussel: *the* status api

Also, can this be checked via the status api.  If not, remove that part of the message.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
sudarsanm94: Yes. Status first verifies whether Presto is installed.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
sudarsanm94: Yes. If the presto is not installed the status displays a message saying "presto is not installed"

rschlussel: Thanks for the clarification!  Please the message to say "using *the* status api"  not "using status api".

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
rschlussel: 1. say "using *the* status api" instead of "using status api"   
2. say "to make sure that installation succeeded"  not "to make sure that install is successful.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
rschlussel: please address this comment.

PR #27, opened 2017-07-26
Add entity to all responses

ShanBai6:


presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/APIFileHandler.java
```diff
@@ -77,15 +79,15 @@ public Response replaceFileFromURL(String path, String url)
         try {
             AgentFileUtils.replaceFile(Paths.get(baseDir.toString(), path), url);
             LOGGER.debug("Successfully replaced file '%s' with url '%s'", path, url);
-            return Response.status(Status.ACCEPTED).build();
+            return Response.status(Status.ACCEPTED).entity("Successfully replaced file with url").build();
```
jirassimok: `ACCEPTED` responses generally should say that they acknowledge the request, not that it worked.

Though here the problem is that we're sending back `ACCEPTED` instead of something more appropriate for a synchronous request, like `OK`, in which case the body might be something like "Set property '%s' to '%s'".

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/APIFileHandler.java
```diff
@@ -132,11 +134,11 @@ public Response getFileProperty(String path, String property)
         }
         catch (NoSuchElementException | FileNotFoundException e) {
             LOGGER.error(e, "File '%s' not found", path);
-            return Response.status(Status.NOT_FOUND).build();
+            return Response.status(Status.NOT_FOUND).entity("File not found").build();
         }
         catch (IOException e) {
             LOGGER.error(e, "Failed to process file '%s'", path);
-            return Response.status(Status.INTERNAL_SERVER_ERROR).build();
+            return Response.status(Status.INTERNAL_SERVER_ERROR).entity("Failed to process file").build();
```
jirassimok: `property of file` → `file property`

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/APIFileHandler.java
```diff
@@ -94,15 +96,15 @@ public Response updateProperty(String path, String property, String value)
         try {
             AgentFileUtils.updateProperty(Paths.get(baseDir.toString(), path), property, value);
             LOGGER.debug("Successfully updated property '%s' of file '%s' to '%s'", property, path, value);
-            return Response.status(Status.ACCEPTED).build();
+            return Response.status(Status.ACCEPTED).entity("Successfully updated property of file").build();
```
jirassimok: `property of file` → `file property`

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
```diff
@@ -212,7 +212,7 @@ private static Response deleteLogRange(Path filePath, Instant end)
         }
 
         // TODO: Use Status.ACCEPTED for asynchronous response
-        return Response.status(Status.NO_CONTENT).build();
+        return Response.status(Status.NO_CONTENT).entity("Logs successfully deleted").build();
```
jirassimok: ditto

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/APIFileHandler.java
jirassimok: `ACCEPTED` responses generally should say that they acknowledge the request, not that it worked.

Though here the problem is that we're sending back `ACCEPTED` instead of something more appropriate for a synchronous request, like `OK`, in which case the body might be something like "Set property '{file}/{property}' to '{value}'". This comes up for quite a few of the responses changed in this PR.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
```diff
@@ -161,7 +161,7 @@ public static Response deleteLogs(String filename, JaxrsParameter<Instant> endDa
             try {
                 Files.write(filePath, new byte[0], StandardOpenOption.TRUNCATE_EXISTING);
                 // TODO: Use Status.ACCEPTED for asynchronous response
-                return Response.status(Status.NO_CONTENT).build();
+                return Response.status(Status.NO_CONTENT).entity("Logs successfully deleted").build();
```
jirassimok: Don't add entities for `NO_CONTENT` responses.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/LogsHandler.java
jirassimok: ditto

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/APIFileHandler.java
jirassimok: `property of file` → `file property`

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/APIFileHandler.java
```diff
@@ -64,62 +68,73 @@ public Response getFile(String path)
...
 
     public Response replaceFileFromURL(String path, String url)
     {
         try {
             AgentFileUtils.replaceFile(Paths.get(baseDir.toString(), path), url);
-            LOGGER.debug("Successfully replaced file '%s' with url '%s'", path, url);
-            return Response.status(Status.ACCEPTED).build();
+            LOGGER.debug("Replacing file '%s' with url '%s' in progress", path, url);
+            return Response.status(Status.ACCEPTED)
+                    .entity("Replacing file with url in progress").build();
```
jirassimok: This operation is not asynchronous; we should be sending back an `OK` with a message that the file was replaced.

ShanBai6: We could change that later, is not relevant to the topic of this PR

jirassimok: But then the newly-added message is inaccurate, which is bad, and relevant to the PR.

PR #29, opened 2017-07-28
Add method to get Presto version

sudarsanm94:


presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -61,22 +67,75 @@ private int execute()
...
+                commandResult.setOutput(outputThread.get(timeout, SECONDS));
+            }
+            catch (TimeoutException e) {
                 process.destroyForcibly();
                 throw new PrestoManagerException(format("Command timed out: %s", commandString));
             }
-            return process.exitValue();
+            finally {
+                executor.shutdownNow();
+            }
```
jirassimok: The preceding blocks need to handle the end of the `executor`, `outputThread`, and `process` more carefully. I can't tell if all possible (or even likely) outcomes are safely handled here.

I recommend looking into `ListenableFuture` and `Futures::addCallback` (both from `com.google.common.util.concurrent`).

On a related note, I just found [this Stack Overflow answer](https://stackoverflow.com/a/9832633), which looks like it could help with the "blocking reader" problem you had. It would also let you log out partial output for a command that takes too long to finish.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -61,22 +67,75 @@ private int execute()
...
+            commandResult.setExitValue(process.exitValue());
+            return commandResult;
         }
-        catch (IOException | InterruptedException e) {
+        catch (IOException | ExecutionException | InterruptedException | IllegalThreadStateException e) {
             throw new PrestoManagerException(format("Error executing command: %s", commandString), e);
         }
     }
+
+    public class ExecuteCallable
```
jirassimok: This class can be replaced with this method, which has the exact body of `ExecuteCallable#call()`:
```java
private getProcessOutput(Process process) { ... }
```
Then you can replace the single instantiation of `ExecuteCallable` with a lambda:
```java
executor.submit(() -> getProcessOutput(process))
```

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
```diff
@@ -61,22 +67,75 @@ private int execute()
...
+            }
+            catch (IOException e) {
+                LOGGER.error("Failed to retrieve the process output", e);
+                return null;
+            }
+            return result;
+        }
+    }
+
+    public class CommandResult
```
jirassimok: This class should be immutable. Given its small size, just replace the setters with a two-argument constructor.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoRpmControlUtils.java
```diff
@@ -138,6 +138,7 @@ public static Response restartUsingRpm()
 
     /**
      * The status api can be used to ascertain whether presto is installed and running.
+     * If Presto is installed, this method would return Presto version
```
jirassimok: Javadocs should be a _statement_ of functionality, not a description.

The summary line should read something like "Get a response representing the status of the managed Presto node". (If more descriptive lines are present, each paragraph (and the summary) should be separated by a line containing only `<p>`.)

All occurences of "would" should be replaced by "will", or better yet, "would return" by "returns".

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoRpmControlUtils.java
```diff
@@ -172,7 +175,7 @@ public static Response statusUsingRpm()
         }
         catch (ProcessingException e) {
             LOGGER.info("Presto is not running");
-            return Response.status(OK).entity(GSON.toJson("Presto is installed but not running"))
+            return Response.status(OK).entity(GSON.toJson(format("Presto version '%s' is installed but not running", prestoVersion)))
```
jirassimok: This is not a useful JSON response. I would expect a JSON response to look something like this:
```json
{
    "running": true,
    "version": "1.43.6",
    "environment": "Tundra",
    "coordinator": false
}
```
You should be able to achieve this by passing a `Map<String, Object>` or something similar to `Gson::toJson`. Don't worry about getting the values' types right; we can fix those when this is switched to Jackson.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -128,7 +128,8 @@ public synchronized Response upgrade(@ApiParam("Url to fetch package") String ur
             LOGGER.error("Invalid url: {}", urlToFetchPackage, e);
             return Response.status(BAD_REQUEST).entity("Invalid url").build();
         }
-        return Response.status(ACCEPTED).entity("Presto is being upgraded.").build();
+        return Response.status(ACCEPTED).entity("Presto is being upgraded. " +
+                "Please check back later using the status api to make sure that upgrade succeeded.").build();
```
jirassimok: To be consistent with the airlift PR, put replace the space at the end of the first string with `\r\n`, and use this message:
> To verify that the upgrade succeeded, check back later using the status API.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/CommandExecutor.java
sudarsanm94: I think the things to worry about are processStream, process itself and the thread.
ProcesStream is closed automatically using the try-with-resources block.
process is destroyed if timeout occurs
and executor is shutdown in the finally block. I think all the cases are handled.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoRpmController.java
```diff
@@ -176,8 +182,11 @@ public Response statusUsingRpm()
...
-            return Response.status(OK).entity(GSON.toJson("Presto is installed but not running"))
-                    .type(APPLICATION_JSON).build();
+            Map<String, Object> statusMap = ImmutableMap.of(
+                    "installed", true,
+                    "running", false,
+                    "version", prestoVersion);
+            return Response.status(OK).entity(GSON.toJson(statusMap)).type(APPLICATION_JSON).build();
         }
         catch (IOException e) {
             LOGGER.error(e, "Failed to get status.");
```
jirassimok: On the line after this, I don't convert the error entity to JSON.

jirassimok: On the line after this, don't convert the error entity to JSON.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoRpmController.java
```diff
@@ -152,12 +156,14 @@ public Response restartUsingRpm()
...
             if (!isInstalled()) {
                 LOGGER.info("Presto is not installed");
-                return Response.status(OK).entity(GSON.toJson("Presto is not installed"))
+                return Response.status(OK).entity(GSON.toJson(ImmutableMap.of("installed", false)))
                         .type(APPLICATION_JSON).build();
             }
+            prestoVersion = getVersion();
         }
         catch (PrestoManagerException e) {
             LOGGER.error(e.getCause(), e.getMessage());
```
jirassimok: On the line after this, don't convert the error entity to JSON (which won't happen until you rebase onto master again).

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUninstaller.java
```diff
@@ -52,22 +53,23 @@ public void runCommand()
...
             return;
         }
         if (executeCommand("service", "presto", "status") == 0) {
-            if (!ignoreErrors) {
+            if (!forceUninstall) {
                 throw new PrestoManagerException("Presto is running");
             }
-            LOGGER.warn("Presto is running, but uninstall will be attempted; Presto will be unstoppable");
+            LOGGER.warn("Presto is running, but uninstall will be attempted");
+            terminatePresto();
```
sudarsanm94: I was indifferent between the options kill or terminate.
Jessica said that terminating presto would be a better option.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUninstaller.java
```diff
@@ -52,22 +53,23 @@ public void runCommand()
...
             LOGGER.warn("Package '%s' is not installed; Process exited with return value: %s", PRESTO_PACKAGE, checkPackageInstalled);
             return;
         }
         if (executeCommand("service", "presto", "status") == 0) {
-            if (!ignoreErrors) {
+            if (!forceUninstall) {
                 throw new PrestoManagerException("Presto is running");
             }
-            LOGGER.warn("Presto is running, but uninstall will be attempted; Presto will be unstoppable");
+            LOGGER.warn("Presto is running; Presto would be forcibly stopped before attempting to uninstall.");
```
jirassimok: `would` → `will` here and in the other error messages, then you're good.

sudarsanm94: Done

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoRpmController.java
```diff
@@ -142,7 +145,8 @@ public Response restartUsingRpm()
 
     /**
      * The status api can be used to ascertain whether presto is installed and running.
-     * If Presto is running, this method would return a JSON file containing the following details
+     * If Presto is installed, this method returns Presto version
```
rschlussel: note that it's json either way

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -125,7 +125,8 @@ public synchronized Response upgrade(@ApiParam("Url to fetch package") String ur
             LOGGER.error(e, "Invalid url: %s", urlToFetchPackage);
             return Response.status(BAD_REQUEST).entity("Invalid url").build();
         }
-        return Response.status(ACCEPTED).entity("Presto is being upgraded.").build();
+        return Response.status(ACCEPTED).entity("Presto is being upgraded.\r\n" +
```
rschlussel: unrelated change.  pull it out into a separate commit

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoRpmController.java
```diff
@@ -152,12 +156,14 @@ public Response restartUsingRpm()
      */
     public Response statusUsingRpm()
     {
+        String prestoVersion;
         try {
             if (!isInstalled()) {
                 LOGGER.info("Presto is not installed");
-                return Response.status(OK).entity(GSON.toJson("Presto is not installed"))
+                return Response.status(OK).entity(GSON.toJson(ImmutableMap.of("installed", false)))
```
rschlussel: unrelated change.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoRpmController.java
```diff
@@ -171,13 +177,21 @@ public Response statusUsingRpm()
                     .request(APPLICATION_JSON).buildGet().invoke();
             JsonParser jsonParser = new JsonParser();
             JsonObject prestoStatus = jsonParser.parse(prestoInfo.readEntity(String.class)).getAsJsonObject();
+            String version = prestoStatus.getAsJsonObject("nodeVersion").get("version").getAsString();
```
rschlussel: 180-185 are unrelated changes.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
sudarsanm94: This message is updated since presto version can be obtained now if presto is not running

PR #30, opened 2017-07-28
Refactor Controller API and Adding the missing APIs

ShanBai6:
1. Add abstract class for methods shared across controller API and refactor the existing ones.
2. Add the remaining controller API and refactor the agent utility classes as needed.


jirassimok:
One specific change you can make:

ApiRequester can not accept null query parameters; there are two steps that should be taken about this:
First, each query parameter should only be added if given by the caller. To facilitate this, this method should be added to AbstractControllerAPI, and used for all of the query parameters that need to be send to the agents:
```java
protected void addQueryParam(ApiRequester request, String name, Object value) {
    if (value != null) {
        request.queryParam(name, value);
    }
}
```

Second, `ApiRequester#queryParam(String, Object...)` should start with this block:
```java
for (Object o : values) {
    requireNonNull(o);
}
```

jirassimok:
The DateParameter stuff should probably be noted more prominently, probably in their own commit, but that would be a bit tricky.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoRpmController.java
```diff
@@ -17,6 +17,7 @@
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
 import com.google.inject.Inject;
+import com.teradata.prestomanager.common.StopType;
```
cawallin: Unnecessary change?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/ControlAPI.java
```diff
@@ -15,7 +15,8 @@
 
 import com.google.inject.Inject;
 import com.teradata.prestomanager.agent.PrestoRpmController;
-import com.teradata.prestomanager.agent.StopType;
+import com.teradata.prestomanager.common.StopType;
+import io.airlift.log.Logger;
```
cawallin: Unnecessary change?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/LogsAPI.java
```diff
@@ -45,20 +40,6 @@
         date = "2017-06-23T09:53:13.549-04:00")
 public class LogsAPI
 {
-    private static final String DEFAULT_DATE = "DEFAULT";
```
cawallin: Ideally, fixing DateParameter should be a separate commit.

presto-manager-common/src/main/java/com/teradata/prestomanager/common/JaxrsParameter.java
```diff
@@ -39,6 +39,8 @@ public JaxrsParameter(String s)
     protected abstract T parseString(String s)
             throws ParseException;
 
+    protected abstract String getString();
```
cawallin: Why getString() and not toString()?

presto-manager-common/src/main/java/com/teradata/prestomanager/common/StopType.java
```diff
@@ -11,7 +11,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.teradata.prestomanager.agent;
```
cawallin: Ideally, this also would be in a separate commit

presto-manager-common/src/main/java/com/teradata/prestomanager/common/JaxrsParameter.java
cawallin: Instead of getString(), consider overriding toString().

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/api/ControllerControlAPI.java
```diff
@@ -0,0 +1,114 @@
...
+            RequestDispatcher requestDispatcher)
+    {
+        super(forwardingClient, requestDispatcher);
+    }
+
+    @POST
+    @Path("/start")
+    @Produces(MediaType.TEXT_PLAIN)
+    public Response startPresto(
+            @QueryParam("scope") String scope,
```
cawallin: How does this handle if the scope and/or nodeId list aren't specified?

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/api/ControllerControlAPI.java
```diff
@@ -0,0 +1,114 @@
...
+    public Response stopPresto(
+            @QueryParam("stopType") @DefaultValue("GRACEFUL") StopType stopType,
+            @QueryParam("scope") String scope,
+            @QueryParam("nodeId") List<UUID> nodeId)
+    {
+        ApiRequester.Builder apiRequester = requesterBuilder(ControllerControlAPI.class)
+                .httpMethod(POST)
+                .pathMethod("stopPresto");
+
+        OptionalQueryParam(apiRequester, "stopType", stopType);
```
cawallin: If this has a default value, it's not optional, right?

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/api/ControllerControlAPI.java
```diff
@@ -0,0 +1,114 @@
...
+        return forwardRequest(scope, apiRequester.build(), nodeId);
+    }
+
+    @POST
+    @Path("/restart")
+    public Response restartPresto(
+            @QueryParam("scope") String scope,
+            @QueryParam("nodeId") List<UUID> nodeId)
+    {
+        ApiRequester apiRequester = requesterBuilder(ControllerControlAPI.class)
```
cawallin: What type of stop does restart do? It seems like (except for the coordinator) you could choose to have a graceful stop here too.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/api/ControllerLogsAPI.java
```diff
@@ -0,0 +1,112 @@
...
+            @QueryParam("scope") String scope,
+            @QueryParam("nodeId") List<UUID> nodeId)
+    {
+        ApiRequester.Builder apiRequester = requesterBuilder(ControllerLogsAPI.class)
+                .httpMethod(GET)
+                .accept(MediaType.TEXT_PLAIN)
+                .pathMethod("getLog")
+                .resolveTemplate("file", file);
+
+        OptionalQueryParam(apiRequester, "from", fromDate.getString());
```
cawallin: from, to, and level don't seem optional

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/api/ControllerLogsAPI.java
```diff
@@ -0,0 +1,112 @@
...
+            @QueryParam("scope") String scope,
+            @QueryParam("nodeId") List<UUID> nodeId)
+    {
+        ApiRequester.Builder apiRequester = requesterBuilder(ControllerLogsAPI.class)
+                .httpMethod(DELETE)
+                .accept(MediaType.TEXT_PLAIN)
+                .pathMethod("deleteLog")
+                .resolveTemplate("file", file);
+
+        OptionalQueryParam(apiRequester, "to", toDate.getString());
```
cawallin: ditto on optional

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/AbstractControllerAPI.java
```diff
@@ -34,6 +34,13 @@ protected AbstractControllerAPI(
         this.requestDispatcher = requestDispatcher;
     }
 
+    protected void OptionalQueryParam(ApiRequester.Builder request, String name, Object value)
```
cawallin: This method needs to start with lower case

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/api/ControllerPackageAPI.java
```diff
@@ -0,0 +1,108 @@
...
+    @Produces({MediaType.TEXT_PLAIN})
+    public Response install(String urlToFetchPackage,
+            @QueryParam("checkDependencies") @DefaultValue("true") boolean checkDependencies,
+            @QueryParam("scope") String scope,
+            @QueryParam("nodeId") List<UUID> nodeId)
+    {
+        ApiRequester.Builder apiRequester = requesterBuilder(ControllerPackageAPI.class)
+                .httpMethod(PUT)
+                .accept(MediaType.TEXT_PLAIN)
+                .entity(Entity.entity(urlToFetchPackage, MediaType.TEXT_PLAIN));
```
cawallin: If urlToFetchPackage can't be null, you should add an explicit null check

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/api/ControllerLogsAPI.java
```diff
@@ -0,0 +1,112 @@
...
+            @QueryParam("level") @DefaultValue("ALL") String level,
+            @QueryParam("n") Integer maxEntries,
+            @QueryParam("scope") String scope,
+            @QueryParam("nodeId") List<UUID> nodeId)
+    {
+        ApiRequester.Builder apiRequester = requesterBuilder(ControllerLogsAPI.class)
+                .httpMethod(GET)
+                .accept(MediaType.TEXT_PLAIN)
+                .pathMethod("getLog")
+                .resolveTemplate("file", file);
```
cawallin: Make sure to do a null check for file, it needs to be non-null.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoRpmController.java
ShanBai6: Since `Controller` already have dependency on `common` module, we move that from agent to common so that we could use it in `Controller` module

presto-manager-common/src/main/java/com/teradata/prestomanager/common/JaxrsParameter.java
jirassimok: `null` is a poor string representation.

It might make more sense to return an `Optional<String>`. Then it might also make sense to change the behavior of `get()` to behave like `Optional::get`.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/api/ControllerControlAPI.java
ShanBai6: It is handled when it calls `forwardRequest` inside `RequestDispatcher`

presto-manager-common/src/main/java/com/teradata/prestomanager/common/StopType.java
ShanBai6: Yes, will do that next time encountering this situation. Already have too many other ones on top of it.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/api/ControllerLogsAPI.java
ShanBai6: It is dealt with inside Agent utility

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/AbstractControllerAPI.java
ShanBai6: Updated

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/api/ControllerPackageAPI.java
ShanBai6: This is also checked inside the agent. Agent will send back `400 BAD REQUEST` in that situation

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/api/ControllerControlAPI.java
ShanBai6: For rpm, restart will always happen no matter of its current state.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/api/ControllerLogsAPI.java
jirassimok: They are. Unfortunately, Jersey doesn't recognize `Optional` in parameters.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/api/ControllerControlAPI.java
jirassimok: This is a bit of a tricky situation.

JAX-RS should map values for any parameter type that has a single-String constructor, but Jersey seems to have a bug where `null` is never passed in, so a missing parameter always has a value of `null`.
So to make `DateParameter` not require null checks, I provide a default that is parsed the same way `null` is supposed to be parsed.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/api/ControllerControlAPI.java
sudarsanm94: Yes.
For rpm restart Presto is automatically stopped if it is running. 
For upgrade and uninstall I terminate presto.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/api/ControllerPackageAPI.java
sudarsanm94: I think its better to have a null check here so that `BAD REQUEST` is not forwarded to the Agent.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/api/ControllerControlAPI.java
jirassimok: Resolved:

We replaced `JaxrsParameter` with Jersey's native parameter conversion API, `ParamConverterProvider`.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoRpmController.java
cawallin: Ah, was it in the same package before so you didn't have to import it?

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/AbstractControllerAPI.java
cawallin: This isn't updated

PR #25, opened 2017-07-26
Switch to Airlift and add Controller server

jirassimok:

PR #31, opened 2017-07-28
Add discovery server and fix controller JSON mapping

jirassimok:
This should be merged after pull request #25.

These two commits don't really belong in the same commit. We need to get a lot of more-or-less unrelated changes through, and pull request #30 depends on both of these commits.

### DIFF
--- a/presto-manager-controller/pom.xml
+++ b/presto-manager-controller/pom.xml
@@ -24,6 +24,12 @@
             <artifactId>presto-manager-common</artifactId>
             <version>${project.version}</version>
         </dependency>
+
+        <dependency>
+            <groupId>io.airlift.discovery</groupId>
+            <artifactId>discovery-server</artifactId>
+            <version>1.28</version>
+        </dependency>
     </dependencies>
 
     <build>
--- a/presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/ControllerServer.java
+++ b/presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/ControllerServer.java
@@ -15,12 +15,15 @@ package com.teradata.prestomanager.controller;
 
 import io.airlift.bootstrap.Bootstrap;
 import io.airlift.discovery.client.DiscoveryModule;
+import io.airlift.discovery.server.DiscoveryServerModule;
 import io.airlift.event.client.HttpEventModule;
 import io.airlift.http.server.HttpServerModule;
 import io.airlift.jaxrs.JaxrsModule;
+import io.airlift.jmx.JmxModule;
 import io.airlift.json.JsonModule;
 import io.airlift.log.Logger;
 import io.airlift.node.NodeModule;
+import org.weakref.jmx.guice.MBeanModule;
 
 public class ControllerServer
 {
@@ -33,16 +36,21 @@ public class ControllerServer
     {
         // TODO: Replace placeholder properties with proper configuration
         System.setProperty("node.environment", "test");
-        System.setProperty("http-server.http.port", "8081");
+        System.setProperty("http-server.http.port", "8088");
         System.setProperty("log.levels-file", "etc/log.properties");
+        System.setProperty("discovery.uri", "http://localhost:"
+                + System.getProperty("http-server.http.port"));
 
         Bootstrap bootstrap = new Bootstrap(
                 new NodeModule(),
                 new DiscoveryModule(),
+                new DiscoveryServerModule(),
+                new HttpEventModule(),
                 new HttpServerModule(),
                 new JsonModule(),
-                new JaxrsModule(true), // requireExplicitBindings = true
-                new HttpEventModule(),
+                new JaxrsModule(true), // requireExplicitBindings = true'
+                new JmxModule(),
+                new MBeanModule(),
                 new ControllerServerModule()
         );
 
--- a/presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/ControllerServerModule.java
+++ b/presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/ControllerServerModule.java
@@ -29,7 +29,6 @@ import javax.ws.rs.client.Client;
 
 import static io.airlift.discovery.client.DiscoveryBinder.discoveryBinder;
 import static io.airlift.jaxrs.JaxrsBinder.jaxrsBinder;
-import static org.weakref.jmx.guice.ExportBinder.newExporter;
 
 public class ControllerServerModule
         extends AbstractConfigurationAwareModule
@@ -40,7 +39,6 @@ public class ControllerServerModule
         binder.disableCircularProxies();
 
         binder.bind(NodeSet.class).in(Scopes.SINGLETON);
-        newExporter(binder).export(NodeSet.class).withGeneratedName();
         binder.bind(RequestDispatcher.class).in(Scopes.SINGLETON);
         binder.bind(Client.class).to(JerseyClient.class).in(Scopes.SINGLETON);
 
@@ -51,6 +49,7 @@ public class ControllerServerModule
     }
 
     @Provides
+    @Singleton
     public JerseyClient jerseyClientProvider()
     {
         return JerseyClientBuilder.createClient();
@@ -59,7 +58,7 @@ public class ControllerServerModule
     @Provides
     @Singleton
     @ForController
-    public ObjectMapper obectMapper()
+    public ObjectMapper objectMapper()
     {
         // TODO: Don't enable the pretty printer ("INDENT_OUTPUT")
         return new ObjectMapper()
--- a/presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/ControllerServerModule.java
+++ b/presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/ControllerServerModule.java
@@ -13,9 +13,12 @@
  */
 package com.teradata.prestomanager.controller;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
 import com.google.inject.Binder;
 import com.google.inject.Provides;
 import com.google.inject.Scopes;
+import com.google.inject.Singleton;
 import com.teradata.prestomanager.controller.api.ControllerConfigAPI;
 import com.teradata.prestomanager.controller.api.ControllerConnectorAPI;
 import io.airlift.configuration.AbstractConfigurationAwareModule;
@@ -51,5 +54,16 @@ public class ControllerServerModule
     public JerseyClient jerseyClientProvider()
     {
         return JerseyClientBuilder.createClient();
+    }
+
+    @Provides
+    @Singleton
+    @ForController
+    public ObjectMapper obectMapper()
+    {
+        // TODO: Don't enable the pretty printer ("INDENT_OUTPUT")
+        return new ObjectMapper()
+                .enable(SerializationFeature.INDENT_OUTPUT)
+                .enable(SerializationFeature.CLOSE_CLOSEABLE);
     }
 }
--- a/presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/ControllerServerModule.java
+++ b/presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/ControllerServerModule.java
@@ -24,7 +24,9 @@ import org.glassfish.jersey.client.JerseyClientBuilder;
 
 import javax.ws.rs.client.Client;
 
+import static io.airlift.discovery.client.DiscoveryBinder.discoveryBinder;
 import static io.airlift.jaxrs.JaxrsBinder.jaxrsBinder;
+import static org.weakref.jmx.guice.ExportBinder.newExporter;
 
 public class ControllerServerModule
         extends AbstractConfigurationAwareModule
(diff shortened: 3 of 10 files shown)


presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/ControllerServerModule.java
```diff
@@ -35,16 +40,30 @@ protected void setup(Binder binder)
         binder.disableCircularProxies();
 
         binder.bind(NodeSet.class).in(Scopes.SINGLETON);
+        newExporter(binder).export(NodeSet.class).withGeneratedName();
```
ShanBai6: What is the purpose of this new exporter, this is more like a question.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/ControllerServerModule.java
```diff
@@ -35,16 +40,30 @@ protected void setup(Binder binder)
...
     @Provides
     public JerseyClient jerseyClientProvider()
     {
         return JerseyClientBuilder.createClient();
     }
+
+    @Provides
+    @Singleton
+    @ForController
+    public ObjectMapper obectMapper()
```
ShanBai6: Misspelling. And it it not used anywhere.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/NodeSet.java
```diff
@@ -13,40 +13,46 @@
...
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
 import static java.util.Objects.requireNonNull;
 
 @ThreadSafe
 public class NodeSet
 {
-    private Set<Agent> nodeSet = ConcurrentHashMap.newKeySet();
+    private static final Logger LOG = Logger.get(NodeSet.class);
```
ShanBai6: We should name the logger to be `LOG` or `LOGGER` always

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/NodeSet.java
```diff
@@ -74,7 +82,40 @@ public void removeAgent(UUID id)
...
+                        properties.get("http"), id.toString());
+                /* TODO: Map this exception to a specific response from the server with...
+                 * ExceptionMapper, or change it to a WebApplicationException containing
+                 * a Response (maybe via a utility class). If using ExceptionMapper, be
+                 * careful not to catch other DiscoveryExceptions (i.e. change this one).
+                 */
+                throw new DiscoveryException("", e);
+            }
+            Agent agent = new Agent(uri, isCoordinator, isWorker, id);
+            setBuilder.add(agent);
```
ShanBai6: Could inline these

presto-manager-controller/pom.xml
```diff
@@ -24,6 +24,12 @@
             <artifactId>presto-manager-common</artifactId>
             <version>${project.version}</version>
         </dependency>
+
+        <dependency>
+            <groupId>io.airlift.discovery</groupId>
+            <artifactId>discovery-server</artifactId>
+            <version>1.28</version>
+        </dependency>
```
jirassimok: We don't need it in the Agent, so I put it here.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/ControllerServerModule.java
jirassimok: That's a good question.

jirassimok: That's a good question.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/ControllerServer.java
```diff
@@ -33,16 +36,21 @@ public static void main(String[] args)
...
                 new DiscoveryModule(),
+                new DiscoveryServerModule(),
+                new HttpEventModule(),
                 new HttpServerModule(),
                 new JsonModule(),
-                new JaxrsModule(true), // requireExplicitBindings = true
-                new HttpEventModule(),
+                new JaxrsModule(true), // requireExplicitBindings = true'
+                new JmxModule(),
+                new MBeanModule(),
```
jirassimok: The order of the modules is sort-of arbitrary. The Node and Discovery modules are first, as in the airlift samples and Presto, then the build-in modules in alphabetical order, then the single custom module.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/ControllerServerModule.java
jirassimok: That should be fixed at some point.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/NodeSet.java
jirassimok: I think it's fine either way.

(I prefer `LOG` because it's shorter, and then you can read "LOG.error" as "log error".)

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/NodeSet.java
jirassimok: I could. I'll add that to one of my upcoming PRs that modifies this file.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/ForController.java
```diff
@@ -0,0 +1,31 @@
...
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Retention(RUNTIME)
+@Target({FIELD, PARAMETER, METHOD})
+@BindingAnnotation
+public @interface ForController
```
sudarsanm94: Can use a better name instead of `ForController`

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/NodeSet.java
sudarsanm94: Then it should be changed everywhere. So it's better to use `LOGGER` for uniformity.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/ForController.java
jirassimok: I tried to think of one, but couldn't, and "For X" seems to be a common naming convention for binding annotations in Airlift.

This particular annotation is used to for injecting something exclusively into the Presto Manager code, when the class being injected already has a binding from airlift.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/NodeSet.java
jirassimok: All right, but I'd prefer to put this change off for a separate commit, where we change all of the `Logger`s to be named `LOGGER`.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/ForController.java
sudarsanm94: Are we planning to add more bindings to this class?
Currently I believe you are planning to bind JSON mapper.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/NodeSet.java
```diff
@@ -13,40 +13,46 @@
...
+    private static final Logger LOG = Logger.get(NodeSet.class);
 
-    @Inject
-    public NodeSet() {}
-
-    public void addAgent(URI uri, boolean isCoordinator, boolean isWorker, UUID id)
-    {
-        nodeSet.add(new Agent(uri, isCoordinator, isWorker, id));
-    }
+    private Set<Agent> nodeSet = ImmutableSet.of();
```
sudarsanm94: Is there a check to see whether `nodeSet` is empty?
Or `nodeSet` being empty is not a problem.

jirassimok: There are potential future situations in which we might operate on `nodeSet` without calling `refresh` first. This protects against those situations at virtually no cost.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/ForController.java
jirassimok: We might in the future, but we do not have any specific plans.

(Though we do have plans to do something similar in the Agent.)

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/ControllerServerModule.java
```diff
@@ -40,11 +44,24 @@ protected void setup(Binder binder)
 
         jaxrsBinder(binder).bind(ControllerConfigAPI.class);
         jaxrsBinder(binder).bind(ControllerConnectorAPI.class);
+
+        discoveryBinder(binder).bindSelector("presto-manager");
     }
 
     @Provides
     public JerseyClient jerseyClientProvider()
```
sudarsanm94: Add `@Singleton`

jirassimok: Will do.

presto-manager-controller/pom.xml
```diff
@@ -24,6 +24,12 @@
             <artifactId>presto-manager-common</artifactId>
             <version>${project.version}</version>
         </dependency>
+
+        <dependency>
+            <groupId>io.airlift.discovery</groupId>
+            <artifactId>discovery-server</artifactId>
+            <version>1.28</version>
```
jirassimok: The discovery server comes from a maven group, so no.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/NodeSet.java
```diff
@@ -74,7 +82,40 @@ public void removeAgent(UUID id)
...
+    {
+        List<ServiceDescriptor> services = serviceSelector.selectAllServices();
+
+        ImmutableSet.Builder<Agent> setBuilder = ImmutableSet.builder();
+
+        for (ServiceDescriptor service : services) {
+            Map<String, String> properties = service.getProperties();
+            // TODO: Make agents start without an ID, and provide one on first discovery
+            UUID id = service.getId();
+            boolean isCoordinator = Boolean.parseBoolean(properties.get("presto-coordinator"));
```
jirassimok: What do I not need within the loop? Every value assigned in the loop is used in the loop and replaced in the next iteration.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/RequestDispatcher.java
```diff
@@ -65,32 +69,39 @@ public Response forwardRequest(String scope, ApiRequester apiRequester, Collecti
...
+                    .entity("Number of coordinator is not 1").build();
         }
 
         List<Response> responseList = new ArrayList<>();
         for (URI uri : uriCollection) {
             responseList.add(apiRequester.send(uri));
         }
+        List<ResponseWrapper> responses = responseList.stream()
+                .map(ResponseWrapper::wrapResponse)
+                .collect(Collectors.toList());
```
jirassimok: There is no existing `Collector` for immutable lists. The easiest solution would be to make an immutable copy of the list after collecting into it, but that seems like a bit much for a list that will be used exactly once. (I could also collect it into an `ImmutableList.Builder`, but that looks awful.)

More importantly, my next pull request rewrites the `for`-loop and stream above entirely. If this section needs immutability, I'd rather add it then than now.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/RequestDispatcher.java
```diff
@@ -40,18 +38,24 @@
...
         this.nodeSet = nodeSet;
+        this.objectMapper = objectMapper;
     }
 
-    public Response forwardRequest(String scope, ApiRequester apiRequester, Collection<UUID> nodeId)
+    public Response forwardRequest(
+            String scope, ApiRequester apiRequester, Collection<UUID> nodeId)
     {
-        if (((scope != null) && (!nodeId.isEmpty())) || (scope == null && nodeId.isEmpty())) {
+        if (((scope != null) && (!nodeId.isEmpty()))
```
jirassimok: Oops, sorry about that.

jirassimok:
@maciejgrzybek I responded to your comments.

As for the squashing of the fixup commits, the only reason that commit is separate is because there are branches that aren't mine off of the earlier commits in this pull requests. If I change those, we're liable to end up with some unnecessarily messy merge conflicts.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/NodeSet.java
maciejgrzybek: Sorry, I misread it. Please disregard my previous comment.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/RequestDispatcher.java
maciejgrzybek: How about `ImmutableList.toImmutableList()` from Guava? I thought you have Guava available in this module.

Anyway, given your next PR is open, we can skip that for now.

jirassimok: Well, that's what I get for looking at StackOverflow instead of the Guavadocs.\

I'll make the rewrite immutable.

PR #35, opened 2017-08-01
Improve request forwarding

jirassimok:
- Allow requests to be forwarded in parallel
- Make NodeSet act like a map to include IDs in responses

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -72,6 +72,14 @@
             <version>2.8.0</version>
         </dependency>
 
+        <!-- TODO: Add the findbugs plugin and use findbugs annotations more
+             @Immutable, @ThreadSafe, and @ -->
+        <dependency>
+            <groupId>com.google.code.findbugs</groupId>
+            <artifactId>annotations</artifactId>
+            <version>3.0.1</version>
+        </dependency>
+
         <dependency>
             <groupId>javax.ws.rs</groupId>
             <artifactId>javax.ws.rs-api</artifactId>
--- a/presto-manager-common/src/main/java/com/teradata/prestomanager/common/ApiRequester.java
+++ b/presto-manager-common/src/main/java/com/teradata/prestomanager/common/ApiRequester.java
@@ -16,6 +16,7 @@ package com.teradata.prestomanager.common;
 import org.eclipse.jetty.http.HttpMethod;
 
 import javax.annotation.Nullable;
+import javax.annotation.concurrent.ThreadSafe;
 import javax.ws.rs.client.Client;
 import javax.ws.rs.client.Entity;
 import javax.ws.rs.client.Invocation;
@@ -30,24 +31,24 @@ import java.util.concurrent.Future;
 
 import static java.util.Objects.requireNonNull;
 
+@ThreadSafe
 public class ApiRequester
 {
-    private final UriBuilder uriBuilder;
+    private final Client client;
+    private final String uriTemplate;
     private final HttpMethod method;
     private final MultivaluedMap<String, Object> headers;
     private final String mediaType;
     @Nullable
-    private Entity entity;
+    private final Entity entity;
 
-    private Client client;
-
-    private ApiRequester(Client client, // TODO: Use AssistedInject to manage this
+    private ApiRequester(Client client,
             UriBuilder uriBuilder, HttpMethod method,
             Entity entity, MultivaluedMap<String, Object> headers,
             String mediaType)
     {
         this.client = requireNonNull(client);
-        this.uriBuilder = requireNonNull(uriBuilder);
+        this.uriTemplate = requireNonNull(uriBuilder).toTemplate();
         this.method = requireNonNull(method);
         this.headers = requireNonNull(headers);
         this.mediaType = requireNonNull(mediaType);
@@ -55,12 +56,28 @@ public class ApiRequester
         this.entity = entity;
     }
 
+    /**
+     * Send this request to the given base URI.
+     * <p>
+     * The path should contain no components besides user-info, host, and port.
+     * <p>
+     * This method is thread-safe.
+     */
+    // TODO: The uri should be validated to not conflict with the uriTemplate
+    // If the uriTemplate contains a path, and the given URI has a path (even
+    // just a single trailing slash, the template's path will be replaced).
     public Response send(URI uri)
     {
         Invocation invocation = createInvocation(uri);
         return invocation.invoke();
     }
 
+    /**
+     * Send this request to the given base URI.
+     * <p>
+     * Identical to {@link #send(URI)}, except the request is sent
+     * asynchronously in a {@link Future}.
+     */
     public Future<Response> sendAsync(URI uri)
     {
         Invocation invocation = createInvocation(uri);
@@ -70,7 +87,7 @@ public class ApiRequester
     private Invocation createInvocation(URI uri)
     {
         Invocation.Builder builder = client
-                .target(uriBuilder.uri(uri).build())
+                .target(UriBuilder.fromUri(uriTemplate).uri(uri).build())
                 .request(mediaType)
                 .headers(headers);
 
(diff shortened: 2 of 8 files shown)


presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/RequestDispatcher.java
```diff
@@ -66,30 +67,31 @@ public Response forwardRequest(
...
-        if (apiScope == ApiScope.COORDINATOR && uriCollection.size() != 1) {
+        if (apiScope == ApiScope.COORDINATOR && uriMap.size() != 1) {
             LOGGER.error("Number of coordinator is not 1");
             return Response.status(INTERNAL_SERVER_ERROR)
                     .entity("Number of coordinator is not 1").build();
         }
 
         // Jackson serializes ArrayLists as JSON arrays
-        ArrayList<ResponseWrapper> responses = uriCollection.parallelStream()
+        Map<UUID, ResponseWrapper> responses = uriMap.entrySet().parallelStream()
```
sudarsanm94: can add `.unordered()`

jirassimok: Set streams are always unordered.

presto-manager-common/src/main/java/com/teradata/prestomanager/common/ApiRequester.java
```diff
@@ -41,26 +41,42 @@
...
+    // If the uriTemplate contains a path, and the given URI has a path (even
+    // just a single trailing slash, the template's path will be replaced).
     public Response send(URI uri)
     {
         Invocation invocation = createInvocation(uri);
         return invocation.invoke();
     }
 
+    /**
+     * Send this reqest to the given base URI.
```
ShanBai6: Typo.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/AgentMap.java
```diff
@@ -0,0 +1,43 @@
...
+    {
+        switch (scope) {
+            case ALL:
+                return getAllUris();
+            case WORKERS:
+                return getWorkerUris();
+            case COORDINATOR:
+                return getCoordinatorUris();
+        }
+        throw new IllegalStateException("Non-exhaustive Enum 'switch' statement");
```
ShanBai6: `IllegalArgument` is more proper

jirassimok: It's an illegal state because it is completely impossible to reach that line (trying to `switch` on `null` throws a `NullPointerException`). The exception is only there because the compiler complains otherwise.

presto-manager-common/src/main/java/com/teradata/prestomanager/common/ApiRequester.java
jirassimok: Fixed.

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/AgentMap.java
ShanBai6: Does not match the documentation. ```Signals that a method has been invoked at an illegal or inappropriate time.``` If just to comfort the compiler, just throw plain exception with explanation will be good.

presto-manager-common/src/main/java/com/teradata/prestomanager/common/ApiRequester.java
```diff
@@ -32,7 +32,7 @@
 
 public class ApiRequester
 {
-    private final UriBuilder uriBuilder;
+    private final String uriTemplate;
```
ShanBai6: What is the reason to change the type here?

jirassimok: I'm storing the URI template now instead of the URI. This allows the `ApiRequester` to safely be used by different threads at the same time.

presto-manager-common/src/main/java/com/teradata/prestomanager/common/ApiRequester.java
```diff
@@ -41,26 +41,42 @@
...
+     * Send this request to the given base URI.
+     * <p>
+     * The path should contain no components besides user-info, host, and port.
+     * <p>
+     * This method is thread-safe.
+     */
+    // TODO: The uri should be validated to not conflict with the uriTemplate
+    // If the uriTemplate contains a path, and the given URI has a path (even
+    // just a single trailing slash, the template's path will be replaced).
     public Response send(URI uri)
```
sudarsanm94: Maybe annotate with `@ThreadSafe`
Same for all the methods which are thread safe

presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/DiscoveryAgentMap.java
```diff
@@ -36,50 +36,60 @@
...
-                return nodeSet.stream().filter(Agent::isCoordinator)
-                        .map(Agent::getUri).collect(Collectors.toList());
-            case WORKERS:
-                return nodeSet.stream().filter(Agent::isWorker)
-                        .map(Agent::getUri).collect(Collectors.toList());
-            default:
-                return null;
-        }
+        return agentSet.stream()
+                .collect(Collectors.toMap(Agent::getId, Agent::getUri));
```
sudarsanm94: You can use `ImmutableMap`  for all these getUris method

jirassimok: Will do.

presto-manager-common/src/main/java/com/teradata/prestomanager/common/ApiRequester.java
jirassimok: I've annotated `ApiRequester` and `RequestDispatcher` as `ThreadSafe` (the annotation only applies to classes).

Maybe I should annotate `ThreadSafe` with the more-specific `Immutable` annotation instead? Or use both?

presto-manager-common/src/main/java/com/teradata/prestomanager/common/ApiRequester.java
```diff
@@ -30,37 +31,53 @@
...
     }
 
+    /**
+     * Send this request to the given base URI.
+     * <p>
+     * The path should contain no components besides user-info, host, and port.
+     * <p>
+     * This method is thread-safe.
+     */
+    // TODO: The uri should be validated to not conflict with the uriTemplate
```
cawallin: How hard is it to add this validation, rather than keeping it as a TODO?

presto-manager-common/src/main/java/com/teradata/prestomanager/common/ApiRequester.java
cawallin: You might as well stick with `@ThreadSafe` on both of these classes

pom.xml
```diff
@@ -72,6 +72,14 @@
             <version>2.8.0</version>
         </dependency>
 
+        <!-- TODO: Add the findbugs plugin and use findbugs annotations more
```
jirassimok: We are. It's the source of `@ThreadSafe`. The TODO is to add the maven plugin that scans the annotations provided by this dependency.

presto-manager-common/src/main/java/com/teradata/prestomanager/common/ApiRequester.java
jirassimok: It's less a matter of fixing this here than of determining where to fix it.

Should we validate URIs as they are discovered? If we validate here, should we only check for conflicting paths? Should we allow the input URI to add new parameters?

PR #37, opened 2017-08-01
Add tarball package type support

sudarsanm94:


presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
```diff
@@ -85,21 +94,130 @@ private static void installUsingRpm(String pathToRpm, boolean checkDependencies)
...
+        addEnvSh();
+        addJvmConfig();
+        addNodeProperties();
+    }
+
+    private static void addConfigProperties()
+            throws PrestoManagerException
+    {
+        Properties properties = new Properties();
+        properties.setProperty("coordinato", "true");
```
ShanBai6: misspelling

ShanBai6: These property should be parameterized and configurable as the same method could be used for deploying on other nodes.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
```diff
@@ -85,21 +94,130 @@ private static void installUsingRpm(String pathToRpm, boolean checkDependencies)
...
+        }
+    }
+
+    private static void addConfigFiles()
+            throws PrestoManagerException
+    {
+        addConfigProperties();
+        addEnvSh();
+        addJvmConfig();
+        addNodeProperties();
```
ShanBai6: Make all these to accept parameter instead of given default values.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
```diff
@@ -85,21 +94,130 @@ private static void installUsingRpm(String pathToRpm, boolean checkDependencies)
...
         }
         catch (IOException e) {
             throw new PrestoManagerException("Failed to add tpch catalog", e);
         }
     }
+
+    private static void installUsingTar(URL urlToFetchPackage)
+            throws PrestoManagerException
+    {
+        preInstallChecks();
```
ShanBai6: Should you do the same checks for RPM?

sudarsanm94: These checks are automatically performed by Presto RPM

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
sudarsanm94: Not needed because. I'm just trying to create the files with default values (same as RPM installation). If user wants to modify these `config API` can be used

sudarsanm94: changed

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
```diff
@@ -85,21 +95,129 @@ private static void installUsingRpm(String pathToRpm, boolean checkDependencies)
...
+    {
+        String jvmConfig = "-server\n" +
+                "-Xmx16G\n" +
+                "-XX:-UseBiasedLocking\n" +
+                "-XX:+UseG1GC\n" +
+                "-XX:+ExplicitGCInvokesConcurrent\n" +
+                "-XX:+HeapDumpOnOutOfMemoryError\n" +
+                "-XX:+UseGCOverheadLimit\n" +
+                "-XX:+ExitOnOutOfMemoryError\n" +
+                "-XX:ReservedCodeCacheSize=512M";
```
jirassimok: We should not be hard-coding this sort of thing.

If we want there to be _any_ default configuration for Presto, we should include that in the configuration for Presto Manager. Better yet, we can provide default Presto configuration by mirroring the Presto configuration directory in one of Presto Manager's directories, and coping its contents after installation.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PackageApiUtils.java
```diff
@@ -69,4 +72,35 @@ public static boolean isRunning()
     {
         return isInstalled() && executeCommand("service", "presto", "status") == 0;
     }
+
+    public static void untarFile(File configTar, Path destinationDir)
+            throws PrestoManagerException
+    {
+        int untarResult = executeCommand("tar", "-xzvf", configTar.getAbsolutePath(), "-C", destinationDir.toAbsolutePath().toString());
+        if (untarResult != 0) {
```
ShanBai6: Inline these.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PackageApiUtils.java
```diff
@@ -69,4 +72,35 @@ public static boolean isRunning()
...
+        }
+    }
+
+    public static void deleteDirContents(Path directory)
+            throws PrestoManagerException
+    {
+        try {
+            deleteDirectoryContents(directory);
+        }
+        catch (IOException e) {
```
ShanBai6: Could be fine-grained. See [Doc](https://google.github.io/guava/releases/21.0/api/docs/com/google/common/io/MoreFiles.html#deleteDirectoryContents-java.nio.file.Path-com.google.common.io.RecursiveDeleteOption...-)

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PackageApiUtils.java
```diff
@@ -19,11 +19,14 @@
...
 
+import static com.google.common.io.MoreFiles.deleteDirectoryContents;
 import static com.teradata.prestomanager.agent.CommandExecutor.executeCommand;
 import static java.io.File.createTempFile;
 import static java.lang.String.format;
 import static java.nio.file.Files.copy;
+import static java.nio.file.Files.createDirectories;
 import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
 
 public class PackageApiUtils
```
jirassimok: This class should be split into two classes, one for RPM, and one for tarballs.

The common method, `fetchFileFromURL` should be removed, and `AgentFileUtils::replaceFile` should be used instead (and renamed to indicate that it downloads a file).

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PackageApiUtils.java
```diff
@@ -69,4 +72,35 @@ public static boolean isRunning()
...
+        }
+    }
+
+    public static void createDir(Path directory)
+            throws PrestoManagerException
+    {
+        try {
+            createDirectories(directory);
+        }
+        catch (IOException e) {
```
ShanBai6: Ditto. explicitly catch `FileAlreadyExist` & `IO` to provide more useful information.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
```diff
@@ -17,24 +17,33 @@
...
 import static java.util.Objects.requireNonNull;
+import static java.util.UUID.randomUUID;
 
 public final class PrestoInstaller
         implements PrestoCommand
 {
     private static final Logger LOGGER = Logger.get(PrestoInstaller.class);
     private static final Path CATALOG_DIR = Paths.get("/etc/presto/catalog");
+    private static final Path TAR_INSTALLATION_DIR = Paths.get("/usr/presto");
+    private static final Path TAR_DATA_DIR = Paths.get("/usr/data");
```
jirassimok: The installation directory should be relative to the user's home directory. I recommend `~/presto-manager/presto/` as the root directory for the installation, containing directories for configuration, logs, and Presto intself.

To get the home directory, use `System.getProperty("user.home")`.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
```diff
@@ -85,21 +95,129 @@ private static void installUsingRpm(String pathToRpm, boolean checkDependencies)
...
             throws PrestoManagerException
     {
         try {
-            if (!isDirectory(CATALOG_DIR)) {
-                createDirectory(CATALOG_DIR);
-                write(CATALOG_DIR.resolve("tpch.properties"), "connector.name=tpch".getBytes());
+            if (!isDirectory(directory)) {
+                createDir(directory);
             }
+            write(directory.resolve("tpch.properties"), "connector.name=tpch".getBytes());
```
ShanBai6: You may want to supply a charset, default would vary from platform to platform

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
```diff
@@ -85,21 +95,129 @@ private static void installUsingRpm(String pathToRpm, boolean checkDependencies)
...
+        finally {
+            if (!tarTempFile.delete()) {
+                LOGGER.warn("Failed to delete the temporary file: %s", tarTempFile.toString());
+            }
+        }
+        createDir(TAR_INSTALLATION_DIR.resolve("etc"));
+        addTpchCatalog(TAR_INSTALLATION_DIR.resolve("etc/catalog"));
+    }
+
+    private static void preInstallChecks()
```
ShanBai6: It is only used once and only contains one line, so probably just put it directly in code.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
```diff
@@ -85,21 +95,129 @@ private static void installUsingRpm(String pathToRpm, boolean checkDependencies)
...
+        }
+        catch (IOException e) {
+            throw new PrestoManagerException("Failed to add config.properties", e);
+        }
+    }
+
+    private static void addNodeProperties()
+            throws PrestoManagerException
+    {
+        // TODO: Add node.server-log-file & node.launcher-log-file
```
ShanBai6: All these add property methods could possibly be refactored as one method that takes a map with name -> value mapping

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
```diff
@@ -85,21 +95,129 @@ private static void installUsingRpm(String pathToRpm, boolean checkDependencies)
...
+        try (OutputStream output = newOutputStream(TAR_INSTALLATION_DIR.resolve("etc/node.properties"), CREATE_NEW)) {
+            properties.store(output, "configurations specific to the node");
+        }
+        catch (IOException e) {
+            throw new PrestoManagerException("Failed to add config.properties", e);
+        }
+    }
+
+    private static void addJvmConfig()
+            throws PrestoManagerException
```
ShanBai6: If this is not in property format, it will cause problems during retrieving & updating

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUninstaller.java
```diff
@@ -83,4 +89,18 @@ private static void uninstallUsingRpm(boolean checkDependencies, boolean forceUn
         }
         LOGGER.debug("Successfully uninstalled Presto");
     }
+
+    public static void uninstallUsingTar()
+            throws PrestoManagerException
+    {
+        // TODO: Stop Presto before uninstall
+        try {
+            deleteRecursively(TAR_INSTALLATION_DIR);
```
ShanBai6: This may cause security problem with recursive deletion if any symbolic link exists. You would like to catch this `InsecureRecursiveDeleteException` and handle it separately. Ditto for the deleteDirectoryContents method you had earlier. This applies to all the places you use this method.

ShanBai6: On another note, these two methods seems to be doing the same thing. You could just use one.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
```diff
@@ -53,19 +62,20 @@ public void runCommand()
     {
         switch (packageType) {
```
jirassimok: Any `PackageType` `switch`-block should be replaced by a single method in each installation-manager class.

For now, have the caller (`PackageAPI`) take the `PackageType` as a constructor parameter, and have a single `switch`-block there in which it determines which installation-manager to use. This can be improved upon with Guice later.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUpgrader.java
```diff
@@ -70,8 +77,7 @@ public void runCommand()
                 }
                 break;
             case TARBALL:
-                // TODO: Add tarball installation
-                throw new UnsupportedOperationException("Tarball upgrade is not supported");
+                upgradeUsingTar(urlToFetchPackage, preserveConfig);
```
jirassimok:
```java
break;
```

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
```diff
@@ -74,12 +76,12 @@ public synchronized Response install(@ApiParam("Url to fetch package") String ur
             return Response.status(BAD_REQUEST).entity("Expected URL in the request body").build();
         }
         try {
-            if (isInstalled()) {
+            if (PACKAGE_TYPE == RPM && isInstalled()) {
```
ShanBai6: What is the point of this check. Final variable will never change and this will always be true?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
```diff
@@ -85,21 +95,129 @@ private static void installUsingRpm(String pathToRpm, boolean checkDependencies)
...
+                createDir(TAR_INSTALLATION_DIR);
+            }
+            untarFile(tarTempFile, TAR_INSTALLATION_DIR);
+        }
+        finally {
+            if (!tarTempFile.delete()) {
+                LOGGER.warn("Failed to delete the temporary file: %s", tarTempFile.toString());
+            }
+        }
+        createDir(TAR_INSTALLATION_DIR.resolve("etc"));
```
jirassimok: The `etc` directory should be stored as a separate constant; if we want to base it off of the install directory, that can be done in the configuration class. Basically, any path you create with `Path::resolve` to which my comment on `addNodeProperties` does not apply should have its own constant.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
```diff
@@ -85,21 +95,129 @@ private static void installUsingRpm(String pathToRpm, boolean checkDependencies)
...
+        }
+    }
+
+    private static void addEnvSh()
+            throws PrestoManagerException
+    {
+        Properties properties = new Properties();
+        properties.setProperty("JAVA8_HOME", System.getProperty("java.home"));
+        try (OutputStream output = newOutputStream(TAR_INSTALLATION_DIR.resolve("etc/env.sh"), CREATE_NEW)) {
+            properties.store(output, "configurations specific to the node");
```
jirassimok: This doesn't seem like the right description for this file.

(And my comment on `addNodeProperties` applies.)

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoUninstaller.java
jirassimok: There should be no risk of `InsecureRecursiveDeleteException` on Linux.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
sudarsanm94: I think this can be improved by Configuration. This should also be done for status

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
sudarsanm94: This is just a test directory which I'm using. This is supposed to be got from the user if user specifies tarball installation.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
jirassimok: That's what I meant by "can be improved upon with Guice".

Eventually, we can have a special installation manager implementation that is configurable with the package type, and selects which of the other implementations to use.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
sudarsanm94: jvm.config won't be in properties format

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
sudarsanm94: In the future `PACKAGE_TYPE` would be configured. So I do this check to determine whether it is RPM or TARBALL installation

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
ShanBai6: Probably create an issue for that after you test this stuff and merge to master

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/PackageAPI.java
ShanBai6: It is `final` though. You could only assign to it once. How would you check if it is changed to TARBALL later.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PackageController.java
```diff
@@ -140,6 +145,37 @@ public Response upgrade(String packageUrl, boolean checkDependencies, boolean pr
                 "To verify that upgrade succeeded, check back later using the status API.").build();
     }
 
+    private void upgradeAsync(URL packageUrl, boolean checkDependencies, boolean preserveConfig)
+            throws PrestoManagerException
+    {
+        File tempPackage = getPackageFile(packageUrl);
+        try {
```
ShanBai6: I would restructure this part. Try block inside another Try block seems ugly

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/TarballController.java
```diff
@@ -0,0 +1,138 @@
...
+        catch (IOException e) {
+            throw new PrestoManagerException(format("Failed to download the file from url : %s", packageUrl.toString()));
+        }
+        return tarTempFile;
+    }
+
+    public void startAsync()
+            throws PrestoManagerException
+    {
+        throw new PrestoManagerException("Unsupported operation");
```
jirassimok: If these are temporary placeholders, they should throw `UnsupportedOperationException("Not implemented")`, and have TODOs to implement.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/TarballController.java
```diff
@@ -0,0 +1,138 @@
...
+    public void restartAsync()
+            throws PrestoManagerException
+    {
+        throw new PrestoManagerException("Unsupported operation");
+    }
+
+    public String getVersion()
+            throws PrestoManagerException
+    {
+        throw new PrestoManagerException("Unsupported operation");
```
jirassimok: If any of these next few functions are not supported by tarball installations, I recommend that the return type be changed to `Optional<current return type>`. For tarball installation, you can return `Optional.empty()`.

Then, when you add them to the JSON, do it like this:

```java
Optional<String> version = getVersion();
version.ifPresent(version -> json.addProperty("version", version));
```
However, I would hold off on this change until after the current problems with JSON parsing are resolved.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/TarballController.java
jirassimok: If these are temporary placeholders, they should throw `UnsupportedOperationException("Not implemented")`, and have TODOs to implement.

Also, the implementations of abstract methods should be annotated with `@Overrides`.
I forgot to say this on `RpmController`, but they should also be annotated there.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/TarballController.java
jirassimok: If any of these next few functions are not supported by tarball installations, I recommend that the return type be changed to `Optional<current return type>`. For tarball installation, you can return `Optional.empty()`.

Then, when you add them to the JSON, do it like this:

```java
Optional<String> version = getVersion();
version.ifPresent(version -> json.addProperty("version", version));
```
However, I would hold off on this change until after the current problems with JSON parsing are resolved.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/TarballController.java
```diff
@@ -0,0 +1,138 @@
...
+import static java.nio.file.Files.createDirectories;
+import static java.nio.file.Files.isDirectory;
+
+public class TarballController
+        extends PackageController
+{
+    private static final Logger LOGGER = Logger.get(TarballController.class);
+
+    private static final Path TAR_INSTALLATION_DIR = Paths.get("/usr/presto");
+    private static final Path TAR_DATA_DIR = Paths.get("/usr/data");
```
jirassimok: I recommend `presto/{presto-installation}` as the tar installation directory, `{tar installation directory}/data` as the data directory, and `{tar installation directory}/config` as the configuration directory.

I would also call these `INSTALL_DIR`, and `DATA_DIR`; the `TAR_` is redundant in the `TarballController` class.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/TarballController.java
```diff
@@ -0,0 +1,138 @@
...
+
+    public void installAsync(URL packageUrl, boolean checkDependencies)
+            throws PrestoManagerException
+    {
+        File tarTempFile = getPackageFile(packageUrl);
```
jirassimok: Please use `java.nio` instead of `java.io` where possible.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigUtils.java
```diff
@@ -59,16 +59,16 @@ public static File storeConfigFiles(Path configDir)
         return tempFile;
     }
 
-    public static void deployConfigFiles(File configTar, Path configDir)
+    public static void untarFile(File src, Path dest)
```
jirassimok: This method should be in `AgentFileUtils`, and both arguments should be `Path`s.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/TarballController.java
```diff
@@ -0,0 +1,138 @@
...
+        }
+        finally {
+            if (!tarTempFile.delete()) {
+                LOGGER.warn("Failed to delete the temporary file: %s", tarTempFile.toString());
+            }
+        }
+        LOGGER.debug("Successfully installed Presto");
+    }
+
+    private void tarInstall(File tarFile)
```
jirassimok: ditto

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/TarballController.java
```diff
@@ -0,0 +1,138 @@
...
+            deleteRecursively(TAR_INSTALLATION_DIR);
+            deleteRecursively(TAR_DATA_DIR);
+        }
+        catch (IOException e) {
+            throw new PrestoManagerException("Failed to uninstall Presto", e);
+        }
+        LOGGER.debug("Successfully uninstalled Presto");
+    }
+
+    public void upgradePackage(File tarFile, boolean checkDependencies)
```
jirassimok: ditto

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/TarballController.java
```diff
@@ -0,0 +1,138 @@
...
+            throws PrestoManagerException
+    {
+        uninstallAsync(checkDependencies);
+        tarInstall(tarFile);
+    }
+
+    public File getPackageFile(URL packageUrl)
```
jirassimok: ditto

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/RpmController.java
```diff
@@ -93,46 +91,15 @@ public void uninstallAsync(boolean checkDependencies)
...
-    }
-
-    private void upgradePackage(String pathToRpm, boolean checkDependencies)
+    public void upgradePackage(File packageFile, boolean checkDependencies)
             throws PrestoManagerException
     {
         int upgradeRpm;
         if (checkDependencies) {
-            upgradeRpm = executeCommand(SUBPROCESS_TIMEOUT, "sudo", "rpm", "-U", pathToRpm);
+            upgradeRpm = executeCommand(SUBPROCESS_TIMEOUT, "sudo", "rpm", "-U", packageFile.toString());
```
jirassimok: This isn't so important, but I keep forgetting to mention it whenever this file changes: it's better to use the long argument names for command line tools when they exist (e.g. `--upgrade` vs `-U`).

jirassimok:
Also, now that I'm looking at code using it, I'm starting to think that `PrestoManager` is a better name than `PrestoController`, regardless of the name of the project. It makes more sense, and I am more likely to get confused between the Presto Manager Controller and `PrestoController` than between Presto Manager and `PrestoManager`, because "Presto Manager" doesn't come up much _inside_ the project.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/TarballController.java
sudarsanm94: It is recommended that data directory be placed outside installation dir. Anyway these things can be configured.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/TarController.java
```diff
@@ -0,0 +1,235 @@
...
+        extends PackageController
+{
+    private static final Logger LOGGER = Logger.get(RpmController.class);
+
+    private final Path configDir;
+    private final Path catalogDir;
+    private final Path dataDir;
+    private final Path pluginDir;
+    private final CommandExecutor executor;
+    private final PrestoConfigDeployer configUtils;
```
sudarsanm94: For `configuration`  this utility class is `package-private`.

rschlussel: I mean that it's not a utility class in the traditional sense.  it's a standard java object that has state and does stuff.  And it should be called by its more descriptive name of configDeployer so people reading the code have a sense of what it does.

sudarsanm94: I changed that. `configUtils` -> `configDeployer` 
I was replying to this comment `Utility classes should have private constructors and consist only of static methods`

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/TarController.java
```diff
@@ -0,0 +1,235 @@
...
+    private static final Path installationDir = Paths.get("~/usr/presto");
+    private static Path launcherScript;
+
+    @Inject
+    TarController(PrestoConfig config, Client client,
+            PrestoInformer informer, CommandExecutor executor,
+            PrestoConfigDeployer configUtils)
+    {
+        super(client, informer);
+        configDir = requireNonNull(config.getConfigurationDirectory());
```
sudarsanm94: `this` is needed only when there is name conflict right? what is the use of `this` here?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/TarController.java
```diff
@@ -0,0 +1,235 @@
...
+        int commandResult = executor.runCommand(launcherScript.toAbsolutePath().toString(), command);
+        if (commandResult != 0) {
+            throw new PrestoManagerException(format("Failed to %s Presto", command), commandResult);
+        }
+    }
+
+    public String getVersion()
+            throws PrestoManagerException
+    {
+        throw new PrestoManagerException("Unsupported operation");
```
sudarsanm94: If Presto is running I get version by querying Presto. This method is called only when Presto is installed and not running.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/TarController.java
```diff
@@ -0,0 +1,273 @@
...
+
+    public void restartAsync()
+            throws PrestoManagerException
+    {
+        executeCommand("restart");
+    }
+
+    private int runLauncherCommand(String command)
+            throws PrestoManagerException
+    {
```
sudarsanm94: No. this `runLauncherCommand` is basically a wrapper for running launcher commands. 
It runs the command like this:
```bin/launcher start CONFIGURATIONS```
CONFIGURATIONS is got from `launcherConfig`

File: presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/TarController.java [hunk 34]
```diff
@@ -0,0 +1,273 @@
...
+                    .filter(path -> isDirectory(path))
+                    .reduce((a, b) -> {
+                        throw new IllegalStateException(format("Multiple directories within `%s`", INSTALLATION_DIR));
+                    })
+                    .get().getFileName();
+            launcherScript = Optional.of(INSTALLATION_DIR.toAbsolutePath().resolve(prestoDir).resolve("bin/launcher"));
+            updateProperty(configDir.toAbsolutePath().resolve("node.properties"), "plugin.dir",
+                    INSTALLATION_DIR.toAbsolutePath().resolve(prestoDir).resolve("plugin").toString());
+            /**
+             * If there are extra spaces in this string, there will be problems
```
sudarsanm94: I split the launcherConfig in `runLauncherCommand` so extra spaces would create empty arguments and `processBuilder` will throw an error while executing the command

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/TarController.java
```diff
@@ -0,0 +1,273 @@
...
+    public Optional<String> getVersion()
+            throws PrestoManagerException
+    {
+        return Optional.empty();
+    }
+
+    public boolean isInstalled()
+            throws PrestoManagerException
+    {
+        return isDirectory(INSTALLATION_DIR) &&
```
sudarsanm94: changed

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/TarController.java
rschlussel: I don't understand how your response relates to changing the error checking for the launcher script

PR #38, opened 2017-08-02
Add config files and connectors during install

sudarsanm94:


presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PackageApiUtils.java
```diff
@@ -58,6 +71,79 @@ public static void checkRpmPackage(String pathToRpm)
...
 
+    /**
+     * Copies all the regular files within a directory to a new location
+     * <p>
+     * <p> The destination directory is created if it does not exist
+     *
+     * @param src Source directory
+     * @param dest Destination directory
+     */
+    public static void copyDir(Path src, Path dest)
```
jirassimok: This function belongs in `AgentFileUtils`.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
```diff
@@ -85,21 +93,56 @@ private static void installUsingRpm(String pathToRpm, boolean checkDependencies)
         if (installRpm != 0) {
             throw new PrestoManagerException("Failed to install Presto", installRpm);
         }
-        addTpchCatalog();
+        addConfigFiles(CONFIG_DIR, DEFAULT_CONFIG_DIR);
+        addConnectors(CATALOG_DIR, DEFAULT_CATALOG_DIR);
```
jirassimok: These could be put in their own function (`addDefaultFiles`?), but this is fine, too.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
```diff
@@ -85,21 +93,56 @@ private static void installUsingRpm(String pathToRpm, boolean checkDependencies)
...
+                write(catalogDir.resolve("tpch.properties"), "connector.name=tpch".getBytes());
             }
         }
-        catch (IOException e) {
+        catch (IOException | PrestoManagerException e) {
             throw new PrestoManagerException("Failed to add tpch catalog", e);
         }
     }
+
+    private static void addConfigFiles(Path configDir, Path defaultConfigDir)
```
jirassimok: I would name this something like `useDefaultConfig`, and it should not take the default configuration as an argument; it should reference the constant directly.

I would also probably have another function called `addDefaultConfig` that does the same thing, but doesn't delete the existing files first. (Call that one here.)

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInstaller.java
```diff
@@ -85,21 +93,56 @@ private static void installUsingRpm(String pathToRpm, boolean checkDependencies)
...
+            else {
+                addWorkerConfig(configDir);
+            }
+        }
+        catch (PrestoManagerException e) {
+            throw new PrestoManagerException("Failed to add config files", e);
+        }
+    }
+
+    private static void addConnectors(Path catalogDir, Path defaultCatalogDir)
```
jirassimok: ditto, but `Config` → `Catalog`

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/WorkerConfig.java
```diff
@@ -0,0 +1,96 @@
...
+import java.nio.file.Paths;
+import java.util.Properties;
+
+import static java.lang.String.format;
+import static java.nio.file.Files.newOutputStream;
+import static java.nio.file.Files.write;
+import static java.nio.file.StandardOpenOption.CREATE_NEW;
+import static java.util.UUID.randomUUID;
+
+public class WorkerConfig
```
jirassimok: This class should be replaced with default configuration files (which should probably be in `etc/defaults`).

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigUtils.java
```diff
@@ -74,4 +74,123 @@ public static void deployConfigFiles(File configTar, Path configDir)
             throw new PrestoManagerException("Failed to deploy config files", untarResult);
         }
     }
+
+    public static void addConfigFiles(Path configDir)
+            throws PrestoManagerException
+    {
+        try {
+            deleteDirectoryContents(configDir);
+            if (containsRegularFiles(DEFAULT_CONFIG_DIR)) {
```
ShanBai6: You should check if it has all the configuration files instead of just checking if there is files inside

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigUtils.java
```diff
@@ -74,4 +74,123 @@ public static void deployConfigFiles(File configTar, Path configDir)
...
+                copyDir(DEFAULT_CONFIG_DIR, configDir);
+                updateProperty(configDir.resolve("node.properties"), "node.id", randomUUID().toString());
```
jirassimok: It would be better if we could have this node ID be included in the install request, so all of the IDs are generated by the Controller.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigUtils.java
```diff
@@ -74,4 +74,123 @@ public static void deployConfigFiles(File configTar, Path configDir)
...
+    private static void addTpchCatalog(Path catalogDir)
+            throws IOException
+    {
+        if (!isDirectory(catalogDir)) {
+            createDirectories(catalogDir);
+            write(catalogDir.resolve("tpch.properties"), "connector.name=tpch".getBytes());
+        }
+    }
+
+    private static void addWorkerConfig(Path configDir)
```
ShanBai6: It is only used once. Probably do not need a separate method

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigUtils.java
jirassimok: It would be better if we could have this node ID be included in the install request, so all of the IDs are generated by the Controller.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/RpmController.java
```diff
@@ -48,7 +49,8 @@ public void installAsync(URL packageUrl, boolean checkDependencies)
             if (installRpm != 0) {
                 throw new PrestoManagerException("Failed to install Presto", installRpm);
             }
-            addTpchCatalog(CATALOG_DIR);
+            addConfigFiles(CONFIG_DIR);
+            addConnectors(CATALOG_DIR);
```
jirassimok: Maybe add `default` to the names of these methods to make it a bit more clear.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigUtils.java
```diff
@@ -74,4 +74,123 @@ public static void deployConfigFiles(File configTar, Path configDir)
...
+                "-Xmx16G\n" +
+                "-XX:-UseBiasedLocking\n" +
+                "-XX:+UseG1GC\n" +
+                "-XX:G1HeapRegionSize=32M\n" +
+                "-XX:+ExplicitGCInvokesConcurrent\n" +
+                "-XX:+HeapDumpOnOutOfMemoryError\n" +
+                "-XX:+UseGCOverheadLimit\n" +
+                "-XX:+ExitOnOutOfMemoryError\n" +
+                "-XX:ReservedCodeCacheSize=512M\n" +
+                "-DHADOOP_USER_NAME=hive";
```
jirassimok: Where do these defaults come from?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigUtils.java
```diff
@@ -74,4 +74,123 @@ public static void deployConfigFiles(File configTar, Path configDir)
...
+        addJvmConfig(configDir);
+        addEnvSh(configDir);
+    }
+
+    private static void addConfigProperties(Path configDir)
+            throws PrestoManagerException
+    {
+        // TODO: Add discovery.uri
+        Properties properties = new Properties();
+        properties.setProperty("coordinator", "false");
```
jirassimok: This is out of the scope of this pull request, but maybe we should add a flag to the installation API for single nodes to determine if it should be the coordinator.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigUtils.java
```diff
@@ -74,4 +74,123 @@ public static void deployConfigFiles(File configTar, Path configDir)
             throw new PrestoManagerException("Failed to deploy config files", untarResult);
         }
     }
+
+    public static void addConfigFiles(Path configDir)
+            throws PrestoManagerException
+    {
+        try {
+            deleteDirectoryContents(configDir);
```
jirassimok: Will this function get called if the user asked to preserve the existing configuration?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigUtils.java
jirassimok: Rather than check for the presence of regular files, you could check if the directory exists at all.

That way, there is a way to prevent the creation of the defaults if the user has other plans.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigUtils.java
sudarsanm94: It'll be nice if there's a separate method which adds all the hard-coded files needed. And also makes it clear that the hard-coded configs are for the worker.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/AgentFileUtils.java
```diff
@@ -137,4 +143,40 @@ public static Path downloadFile(URL url)
...
+            throws IOException
+
+    {
+        List<Path> filePaths = Files.list(src).filter(path -> isRegularFile(path, NOFOLLOW_LINKS))
+                .collect(ImmutableList.toImmutableList());
+        if (!isDirectory(dest)) {
+            createDirectories(dest);
+        }
+        for (Path path : filePaths) {
+            copy(path, dest.resolve(path.getFileName()));
```
sudarsanm94: changed

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigUtils.java
sudarsanm94: No.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigUtils.java
sudarsanm94: what difference does it make whether controller or agent generate the node ID?
Both are going to generate using `randomUUID()`

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigUtils.java
sudarsanm94: I think its not needed. since, discovery will take care of finding the coordinator. By default we just add worker and the user can change it if its needed.
If this feature is absolutely needed, we can do it in the future

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigUtils.java
sudarsanm94: I just took these from `presto-admin`. These can be modified in the future.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigUtils.java
sudarsanm94: I think this is not a good idea., if the Presto config files change in the future.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigUtils.java
```diff
@@ -74,4 +73,123 @@ public static void deployConfigFiles(File configTar, Path configDir)
...
+        createPropertiesFile(properties, configDir.resolve("config.properties"), "single node worker config");
+    }
+
+    private static void addNodeProperties(Path configDir)
+            throws PrestoManagerException
+    {
+        // TODO: Add node.server-log-file & node.launcher-log-file
+        Properties properties = new Properties();
+        properties.setProperty("node.environment", "presto");
+        properties.setProperty("node.id", randomUUID().toString());
```
rschlussel: for upgrade we should keep the previous node.id.    You can leave that as todo

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigUtils.java
```diff
@@ -74,4 +73,123 @@ public static void deployConfigFiles(File configTar, Path configDir)
...
+        catch (IOException e) {
+            throw new PrestoManagerException("Failed to add config files", e);
+        }
+    }
+
+    private static void addEnvSh(Path configDir)
+            throws PrestoManagerException
+    {
+        Properties properties = new Properties();
+        properties.setProperty("JAVA8_HOME", System.getProperty("java.home"));
```
rschlussel: we only care about setting java8_home when java8 is different from JAVA_HOME and not in a standard location, so this doesn't add anything.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigUtils.java
sudarsanm94: If the user specifies `preserveConfig` then node.id is preserved. If not, it'll be replaced.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigUtils.java
sudarsanm94: So don't add this file?

PR #40, opened 2017-08-04
Add configuration capability to application

ShanBai6:
_squashed_

PR #48, opened 2017-08-08
Remove uses of `java.io.File` and rename various methods and classes

jirassimok:
This must be merged _after_ pull request #40.

### DIFF
--- a/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/APIFileHandler.java
+++ b/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/APIFileHandler.java
@@ -19,6 +19,8 @@ import javax.ws.rs.core.Response;
 
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.nio.file.NoSuchFileException;
 import java.nio.file.NotDirectoryException;
 import java.nio.file.Path;
@@ -81,10 +83,15 @@ public class APIFileHandler
     public Response replaceFileFromURL(String path, String url)
     {
         try {
-            AgentFileUtils.replaceFile(Paths.get(baseDir.toString(), path), url);
-            LOGGER.debug("Replacing file '%s' with url '%s' in progress", path, url);
+            AgentFileUtils.replaceFile(Paths.get(baseDir.toString(), path), new URL(url));
+            LOGGER.debug("Replaced file '%s' with file at url '%s'", path, url);
             return Response.status(Status.ACCEPTED)
-                    .entity("Replacing file with url in progress").build();
+                    .entity("File replaced").build();
+        }
+        catch (MalformedURLException | IllegalArgumentException e) {
+            LOGGER.error(e, "Invalid URL: %s", url);
+            return Response.status(Status.BAD_REQUEST)
+                    .entity("Invalid URL").build();
         }
         catch (FileNotFoundException e) {
             LOGGER.error(e, "File '%s' not found", path);
@@ -102,7 +109,7 @@ public class APIFileHandler
     {
         try {
             AgentFileUtils.updateProperty(Paths.get(baseDir.toString(), path), property, value);
-            LOGGER.debug("Updating property '%s' of file '%s' to '%s'", property, path, value);
+            LOGGER.debug("Updated property '%s' in file '%s' to '%s'", property, path, value);
             return Response.status(Status.OK)
                     .entity("Successfully updated the property of file").build();
         }
@@ -122,7 +129,7 @@ public class APIFileHandler
     {
         try {
             AgentFileUtils.deleteFile(Paths.get(baseDir.toString(), path));
-            LOGGER.debug("Deleting file '%s' in progress", path);
+            LOGGER.debug("File deleted", path);
             return Response.status(Status.ACCEPTED)
                     .entity("Deleting file in progress").build();
         }
@@ -161,9 +168,9 @@ public class APIFileHandler
     {
         try {
             AgentFileUtils.removePropertyFromFile(Paths.get(baseDir.toString(), path), property);
-            LOGGER.debug("Deleting property '%s' from file '%s' in progress", property, path);
+            LOGGER.debug("Deleted property '%s' from file '%s'", property, path);
             return Response.status(Status.ACCEPTED)
-                    .entity("Deleting file property in progress").build();
+                    .entity("Deleted property").build();
         }
         catch (NoSuchElementException | FileNotFoundException e) {
             LOGGER.error(e, "File '%s' not found", path);
--- a/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigDeployer.java
+++ b/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigDeployer.java
@@ -15,7 +15,6 @@ package com.teradata.prestomanager.agent;
 
 import com.google.inject.Inject;
 
-import java.io.File;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.file.Path;
@@ -24,9 +23,9 @@ import java.util.Properties;
 import static com.google.common.io.MoreFiles.deleteDirectoryContents;
 import static com.teradata.prestomanager.agent.AgentFileUtils.copyDir;
 import static com.teradata.prestomanager.agent.AgentFileUtils.updateProperty;
-import static java.io.File.createTempFile;
 import static java.lang.String.format;
 import static java.nio.file.Files.createDirectories;
+import static java.nio.file.Files.createTempFile;
 import static java.nio.file.Files.isDirectory;
 import static java.nio.file.Files.newOutputStream;
 import static java.nio.file.Files.write;
@@ -49,10 +48,20 @@ public class PrestoConfigDeployer
         this.executor = requireNonNull(executor);
     }
 
-    public File storeConfigFiles(Path configDir)
+    /**
+     * Backup the given directory in a temporary file.
+     * <p>
+     * The only guarantee about the returned {@link Path} is that it may be
+     * restored with {@link #restoreDirectory(Path, Path)}.
+     * <p>
+     * The backup file may be created in the system's temporary file directory.
+     *
+     * @return The location of the backup file.
+     */
+    public Path backupDirectory(Path directory)
             throws PrestoManagerException
     {
-        File tempFile;
+        Path tempFile;
         try {
             tempFile = createTempFile("PrestoConfigs", ".tar.gz");
         }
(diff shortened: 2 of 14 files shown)


presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigUtils.java
```diff
@@ -68,7 +70,10 @@ public File storeConfigFiles(Path configDir)
         return tempFile;
     }
 
-    public void deployConfigFiles(File configTar, Path configDir)
+    /**
+     * Untar the given tarball into the given directory
+     */
+    public void deployConfigFiles(Path configTar, Path configDir)
```
sudarsanm94: Might as well change this method name to `untarFile` or something like that

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigUtils.java
```diff
@@ -78,26 +83,35 @@ public void deployConfigFiles(File configTar, Path configDir)
...
             throw new PrestoManagerException("Failed to deploy config files", untarResult);
         }
     }
 
-    public void addConfigFiles(Path configDir, Path dataDir, Path pluginDir)
+    /**
+     * Add default Presto configuration in the given directories
+     * <p>
+     * If {@link #defaultConfig} exists, its contents will be used for
+     * the configuration directory. Otherwise, non-configurable default
```
sudarsanm94: Also, state that its worker config files.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigUtils.java
```diff
@@ -106,34 +120,43 @@ public void addConfigFiles(Path configDir, Path dataDir, Path pluginDir)
...
-            throw new PrestoManagerException(format("Failed to create properties file: %s", path.getFileName()), e);
-        }
-    }
-
-    public void addConnectors(Path catalogDir)
+    /**
+     * Add default Presto connector configuration in the given directory
+     * <p>
+     * If {@link #defaultCatalog} exists, its contents will be used for
+     * the configuration directory. Otherwise, non-configurable default
```
sudarsanm94: Only tpch connector will be added if defaultCatalog does not exist

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/APIFileHandler.java
```diff
@@ -81,10 +84,16 @@ public Response getFile(String path)
     public Response replaceFileFromURL(String path, String url)
     {
         try {
-            AgentFileUtils.replaceFile(Paths.get(baseDir.toString(), path), url);
-            LOGGER.debug("Replacing file '%s' with url '%s' in progress", path, url);
+            AgentFileUtils.replaceFile(Paths.get(baseDir.toString(), path),
+                    new URI(url).toURL());
+            LOGGER.debug("Replaced file '%s' with file at url '%s'", path, url);
             return Response.status(Status.ACCEPTED)
```
ShanBai6: This would take some time. Either send back OK or change the log message

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/APIFileHandler.java
```diff
@@ -81,10 +84,16 @@ public Response getFile(String path)
     public Response replaceFileFromURL(String path, String url)
     {
         try {
-            AgentFileUtils.replaceFile(Paths.get(baseDir.toString(), path), url);
-            LOGGER.debug("Replacing file '%s' with url '%s' in progress", path, url);
+            AgentFileUtils.replaceFile(Paths.get(baseDir.toString(), path),
+                    new URI(url).toURL());
```
ShanBai6: What is the point of converting back and forth?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/APIFileHandler.java
ShanBai6: Ditto for below. I have done refactoring once for all before to make them either return 200/202 then have the log message to comply with them.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/APIFileHandler.java
jirassimok: Not sure; fixed.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigDeployer.java
```diff
@@ -78,25 +83,34 @@ public void deployConfigFiles(File configTar, Path configDir)
...
+     * configuration files will be created.
+     */
+    public void writeConfigFiles(Path configDir, Path dataDir, Path pluginDir)
             throws PrestoManagerException
     {
         try {
             if (isDirectory(defaultConfig)) {
                 copyDir(defaultConfig, configDir);
-                updateProperty(configDir.resolve("node.properties"), "node.id", randomUUID().toString());
+                // TODO: Include this in install requests
```
sudarsanm94: State why do you need this TODO. Also make TODO more clear `Include node.id in the install requests`

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigDeployer.java
```diff
@@ -78,25 +83,34 @@ public void deployConfigFiles(File configTar, Path configDir)
...
 
-    public void addConfigFiles(Path configDir, Path dataDir, Path pluginDir)
+    /**
+     * Add default Presto configuration in the given directories
+     * <p>
+     * If {@link #defaultConfig} exists, its contents will be used for
+     * the configuration directory. Otherwise, non-configurable default
+     * configuration files will be created.
+     */
+    public void writeConfigFiles(Path configDir, Path dataDir, Path pluginDir)
```
sudarsanm94: This should be `addConfigFiles`. `writeConfigFiles` is misleading

presto-manager-common/src/main/java/com/teradata/prestomanager/common/ApiRequester.java
```diff
@@ -48,7 +48,7 @@ private ApiRequester(Client client,
             String mediaType)
     {
         this.client = requireNonNull(client);
-        this.uriTemplate = requireNonNull(uriBuilder).toTemplate();
+        this.uriTemplate = new UriCopyBuilder(requireNonNull(uriBuilder));
```
ShanBai6: What is the use of this CopyBuilder?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigDeployer.java
```diff
@@ -105,34 +119,43 @@ public void addConfigFiles(Path configDir, Path dataDir, Path pluginDir)
...
-
-    public void addConnectors(Path catalogDir)
+    /**
+     * Add default Presto connector configuration in the given directory
+     * <p>
+     * If {@link #defaultCatalog} exists, its contents will be used for
+     * the configuration directory. Otherwise, non-configurable default
+     * connector files will be created.
+     */
+    public void writeConnectorFiles(Path catalogDir)
```
sudarsanm94: should be `addConnectors`

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigDeployer.java
```diff
@@ -105,34 +119,43 @@ public void addConfigFiles(Path configDir, Path dataDir, Path pluginDir)
...
+                crateTpchCatalog(catalogDir);
             }
         }
         catch (IOException e) {
             throw new PrestoManagerException("Failed to add connectors", e);
         }
     }
 
-    private static void addTpchCatalog(Path catalogDir)
+    private static void writePropertiesFile(Properties properties, Path path, String comments)
```
sudarsanm94: `createPropertiesFile`

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/ConfigAPI.java
```diff
@@ -41,12 +41,12 @@
 @Singleton
 public final class ConfigAPI
 {
-    private final APIFileHandler apiFileHandler;
+    private final ApiFileHandler apiFileHandler;
```
sudarsanm94: If you are changing `API` -> `Api` change it in all the files ConfigAPI, PackageAPI, ..

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigDeployer.java
```diff
@@ -49,26 +48,32 @@
...
             throw new PrestoManagerException("Failed to tar config files.", tarResult);
         }
         return tempFile;
     }
 
-    public void deployConfigFiles(File configTar, Path configDir)
+    /**
+     * Untar the given tarball into the given directory
+     */
+    public void deployConfigFiles(Path configTar, Path configDir)
```
jirassimok: I've renamed this to `restoreDirectory`.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigDeployer.java
```diff
@@ -49,26 +48,32 @@
         this.executor = requireNonNull(executor);
     }
 
-    public File storeConfigFiles(Path configDir)
+    /**
+     * Save the given directory in a tarball
+     */
+    public Path storeConfigFiles(Path configDir)
```
jirassimok: I've renamed it to `backupDirectory`.

This way, if we decide to change it to not use `tar`, it will still make sense.

I've also added documentation.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigDeployer.java
jirassimok: I believe that `add` is more misleading, because if the directory already exists, the first thing it does is _delete_ its contents; a function called `add` shouldn't be removing files.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigDeployer.java
jirassimok: If we include node ID in install requests, we can more effectively guarantee the uniqueness of the IDs. Currently, there is a change (albeit a very low change) that two Agents will generate the same node ID.

If they are all generated by the Controller, the IDs are effectively guaranteed to be unique.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigDeployer.java
jirassimok: See my comment at `writeConfigFiles`, and the names of these two functions should be consistent.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigDeployer.java
jirassimok: How about `storeProperties`, because it's just a wrapper around `Properties::store`?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/ConfigAPI.java
jirassimok: I only did it here because it's at the beginning of the name, where it's a lot worse to have an acronym.

I let the other ones because the correct name for those is really `Resource`, as in Presto, and that sort of widespread change is much more likely to cause problems than this one.

presto-manager-common/src/main/java/com/teradata/prestomanager/common/ApiRequester.java
jirassimok: It restricts the valid operations on the template to `template.clone().uri(URI).build()`, so that the `ApiRequester`'s template is guaranteed to be immutable and thread-safe to use.

I could change it to just call `clone` and name it something like `UriTemplateProvider`. Then `ApiRequester` could still be internally thread-safe, but allow later operations to use other URI changes. (I can't think of a case where that would be a good idea, but it would also make the intent a bit clearer.)

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigDeployer.java
sudarsanm94: `deployConfigFiles` is much better

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigDeployer.java
sudarsanm94: `deployConnectors`

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigDeployer.java
sudarsanm94: This method actually creates new properties file so `createPropertiesFile` is better

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/api/ConfigAPI.java
sudarsanm94: Add this one as TODO then

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigDeployer.java
```diff
@@ -49,54 +48,78 @@
...
     }
 
-    public File storeConfigFiles(Path configDir)
+    /**
+     * Backup the given directory in a temporary file.
+     * <p>
+     * The only guarantee about the returned {@link Path} is that it may be
+     * restored with {@link #restoreDirectory(Path, Path)}.
+     * <p>
+     * The backup file may be created in the system's temporary file directory.
```
sudarsanm94: `will` be created

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigDeployer.java
```diff
@@ -49,54 +48,78 @@
...
 
-    public void addConfigFiles(Path configDir, Path dataDir, Path pluginDir)
+    /**
+     * Add default Presto configuration in the given directories
+     * <p>
+     * If {@link #defaultConfig} exists, its contents will be used for
+     * the configuration directory. Otherwise, non-configurable default
+     * configuration files will be created.
+     */
+    public void deployDefaultConfig(Path configDir, Path dataDir, Path pluginDir)
```
sudarsanm94: `deployDefaultConfigs`

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigDeployer.java
jirassimok: Using "may" allows us to change that in the future if we want. This way, nobody programming around this should make any assumptions about the backup's location.

PR #56, opened 2017-08-11
Improve efficiency of building URIs in ApiRequester

jirassimok:
This shouldn't conflict with anything.

### DIFF
--- a/presto-manager-common/src/main/java/com/teradata/prestomanager/common/ApiRequester.java
+++ b/presto-manager-common/src/main/java/com/teradata/prestomanager/common/ApiRequester.java
@@ -35,7 +35,7 @@ import static java.util.Objects.requireNonNull;
 public class ApiRequester
 {
     private final Client client;
-    private final String uriTemplate;
+    private final UriCopyBuilder uriTemplate;
     private final HttpMethod method;
     private final MultivaluedMap<String, Object> headers;
     private final String mediaType;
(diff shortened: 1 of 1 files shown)


presto-manager-common/src/main/java/com/teradata/prestomanager/common/ApiRequester.java
```diff
@@ -48,7 +48,7 @@ private ApiRequester(Client client,
             String mediaType)
     {
         this.client = requireNonNull(client);
-        this.uriTemplate = requireNonNull(uriBuilder).toTemplate();
+        this.uriTemplate = new UriCopyBuilder(requireNonNull(uriBuilder));
```
maciejgrzybek: It's not needed because the c-tor of `UriCopyBuilder` checks for nullness already.

presto-manager-common/src/main/java/com/teradata/prestomanager/common/ApiRequester.java
```diff
@@ -102,6 +102,28 @@ private Invocation createInvocation(URI uri)
         return invocation;
     }
 
+    private static class UriCopyBuilder
+    {
+        private final UriBuilder template;
+
+        private UriCopyBuilder(UriBuilder template)
+        {
+            this.template = requireNonNull(template);
```
maciejgrzybek: The message for `requireNonNull` would help debug the problem in case null was passed here.

presto-manager-common/src/main/java/com/teradata/prestomanager/common/ApiRequester.java
sudarsanm94: Changed

presto-manager-common/src/main/java/com/teradata/prestomanager/common/ApiRequester.java
sudarsanm94: Added a message

PR #57, opened 2017-08-11
Add support for deletion of a certain property

ShanBai6:
Support deletion of a certain property of configuration/connector files. This would help the user to recover if they accidentally insert an irrelevant property which would cause failure to start Presto.

### DIFF
--- a/presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/api/ControllerConfigAPI.java
+++ b/presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/api/ControllerConfigAPI.java
@@ -191,4 +191,27 @@ public class ControllerConfigAPI
 
         return forwardRequest(scope, apiRequester, nodeId);
     }
+
+    @DELETE
+    @Path("/{file}/{property}")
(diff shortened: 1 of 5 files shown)


presto-manager-controller/src/main/java/com/teradata/prestomanager/controller/api/ControllerConfigAPI.java
```diff
@@ -191,4 +191,27 @@ public Response deleteConfigFile(
 
         return forwardRequest(scope, apiRequester, nodeId);
     }
+
+    @DELETE
+    @Path("/{file}/{property}")
+    @ApiOperation(value = "Delete a certain property of configuration file")
+    @ApiResponses(value = {
+            @ApiResponse(code = 207, message = "Multiple responses available"),
```
maciejgrzybek: Why multiple responses are available for deleting a single property?

sudarsanm94: Since the same request could be sent to multiple Presto-Manager agents.

PR #62, opened 2017-08-15
Refactor configuration

sudarsanm94:
Fix bugs
Modify PrestoConfigDeployer
Refactor configuration

### DIFF
--- a/etc/agent.properties
+++ b/etc/agent.properties
@@ -22,12 +22,12 @@ log.levels-file       = etc/log.properties
 # configuration file's http-server.http.port property.
 discovery.uri         = http://localhost:8088
 
-# What type of Presto isntallation is being managed.
+# What type of Presto installation is being managed.
 # Valid values: rpm, tarball
 packaging             = rpm
 
 # These properties determine where Presto Manager will look for
-# default onfiguration and connector files for a new Presto
+# default configuration and connector files for a new Presto
 # installation. If these directories are not present, hard-coded
 # defaults will be used.
 # To create an installation with no configuration, create these
@@ -78,11 +78,23 @@ log-entry.default = 0000-01-01T00:00:00.000+0000\
 #### Presto Information #############################################
 
 # These configuration properties tell Presto Manager where to look
-# for Presto's files.
+# for Presto's files. The paths can be absolute or relative to the
+# Presto-Manager directory.
+# The below values should match the corresponding properties in
+# `node.properties` file present in `defaults.config-dir`.
 
-presto.config-dir     = /etc/presto
-presto.catalog-dir    = /etc/presto/catalog
-presto.data-dir       = /var/lib/presto/presto/data
-presto.plugin-dir     = /usr/lib/presto/lib/plugin
-presto.log-dir        = /var/log/presto
-presto.launcher       = /usr/lib/presto/bin/launcher
+# for RPM installation
+
+presto.catalog-dir      = /etc/presto/catalog
+presto.data-dir         = /var/lib/presto/data
+presto.log-dir          = /var/log/presto
+
+# for TARBALL installation
+# presto.launcher-Properties is optional
+
+# presto.installation-dir = /usr/presto
+# presto.config-dir       = /etc/presto
+# presto.catalog-dir      = /etc/presto/catalog
+# presto.data-dir         = /var/lib/presto/presto/data
+# presto.log-dir          = /var/log/presto
+# presto.launcher-Properties         = /usr/lib/presto/bin/launcher
--- a/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfig.java
+++ b/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfig.java
@@ -13,98 +13,28 @@
  */
 package com.teradata.prestomanager.agent;
 
-import io.airlift.configuration.Config;
-
 import javax.validation.constraints.NotNull;
 
 import java.nio.file.Path;
-import java.nio.file.Paths;
+import java.util.Optional;
 
-public class PrestoConfig
+public interface PrestoConfig
 {
-    private Path configurationDirectory;
-    private Path catalogDirectory;
-    private Path dataDirectory;
-    private Path pluginDirectory;
-    private Path logDirectory;
-    private Path configProperties;
-    private Path launcherScript;
-
-    @Config("config-dir")
-    public PrestoConfig setConfigurationDirectory(String path)
-    {
-        configurationDirectory = Paths.get(path);
-        return this;
-    }
+    @NotNull
+    Path getInstallationDirectory();
 
     @NotNull
-    public Path getConfigurationDirectory()
-    {
-        return configurationDirectory;
-    }
-
-    @Config("data-dir")
-    public PrestoConfig setDataDirectory(String path)
-    {
-        dataDirectory = Paths.get(path);
-        return this;
-    }
+    Path getConfigDirectory();
 
     @NotNull
-    public Path getDataDirectory()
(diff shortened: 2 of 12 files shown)


rschlussel:
Haven't reviewed, but it looks like this commit: "Add force upgrade option to controller; Fix bugs"  should be split in two

rschlussel:
I'd also split the other commit into one that extracts an interface and makes corresponding changes for the RPM config, and one that adds support for tarball configuration

sudarsanm94:
Made all the changes

rschlussel:
Modify PrestoConfigDeployer needs a more descriptive commit message.
e.g. "Enable use of non-default config directory paths" (if that's what this does).

rschlussel:
The commits `Add force upgrade option` and `Fix bugs` look good.  Though instead of`Fix bugs` I'd call it`Minor cleanup of PackageController` because you would usually want to know by the commit message what bugs have been fixed.

sudarsanm94:
If you look at the description for the `Modify PrestoConfigDeployer` commit I have explained what this commit does

rschlussel:
You should be able to know what the commit message does from the one line title.  This is a great guide on the topic https://chris.beams.io/posts/git-commit/

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/TarController.java
```diff
@@ -180,6 +181,11 @@ public void upgradeAsync(URL packageUrl, boolean checkDependencies, boolean pres
                     tarInstall(tempPackage);
                     postInstall();
                     configDeployer.restoreDirectory(tempConfigDir, configDir);
+                    updateProperty(configDir.toAbsolutePath().resolve("node.properties"), "plugin.dir",
```
rschlussel: I don't follow.  If configs are preserved why was this wrong in the first place?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigDeployer.java
```diff
@@ -115,11 +115,21 @@ public void deployDefaultConfig(Path configDir, Path dataDir, Path pluginDir)
                 // TODO: Include node.id in install requests
                 updateProperty(configDir.resolve("node.properties"),
                         "node.id", randomUUID().toString());
+                /**
+                 * To prevent conflicts between Presto-Manager configs and
+                 * Presto configs present in {@link #defaultConfig} directory
+                 */
+                Path nodePropertiesPath = configDir.toAbsolutePath().resolve("node.properties");
```
rschlussel: I don't understand why this section is needed.  Why do you update properties for the deploying default configs?  shouldn't that follow defaults?

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfig.java
```diff
@@ -13,98 +13,28 @@
...
-    private Path launcherScript;
-
-    @Config("config-dir")
-    public PrestoConfig setConfigurationDirectory(String path)
-    {
-        configurationDirectory = Paths.get(path);
-        return this;
-    }
-
     @NotNull
```
sudarsanm94: Changed

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/TarController.java
sudarsanm94: Even though the configs are preserved during upgrade, the `pluginDir` changes because `pluginDir` exists within the `installationDir` and `installationDir` changes after upgrade.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigDeployer.java
sudarsanm94: Let's say user provides `configDir` location as `/etc/presto` in the Presto-Manager config files.
But in the default configs for Presto the `configDir` is `/usr/presto`. This would cause a conflict.
I'm updating these variables to prevent these conflicts.

sudarsanm94:
Refactor commit can't be split into two because the split halves doesn't work without the other.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfig.java
rschlussel: oh right we talked about this

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigDeployer.java
rschlussel: is it possible for presto-manager to get this info from the provided configs instead of as a separate property?  because it would be much better for there to be a single source of truth.

etc/agent.properties
```diff
@@ -78,14 +78,23 @@ log-entry.default = 0000-01-01T00:00:00.000+0000\
...
+# for RPM installation
+
+presto.catalog-dir      = /etc/presto/catalog
+presto.data-dir         = /var/lib/presto/data
+presto.log-dir          = /var/log/presto
+
+# for TARBALL installation
+# presto.launcher-Properties is optional
+
+# presto.installation-dir = /usr/presto
```
rschlussel: Are the commented out values supposed to be for tarball installation?  those values don't look right.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigDeployer.java
sudarsanm94: So the following three files are needed for Presto to start:
`jvm.config`, `config.properties`, `node.properties`
Most of the configurable stuff in `node.properties` can be obtained from presto-manager configuration.
But for `jvm.config` & `config.properties` need to be specified by the user.
I think this can be upgraded in the future as an enhancement

etc/agent.properties
sudarsanm94: yes.
These are some default values which I added. The user can specify anything.

rschlussel: but they're misleading because the directories for tar ball install should all be inside the installation directory

sudarsanm94: No they need not be.
Installation dir is where Presto files will be untarred. The config, catalog and other folders can be anywhere

rschlussel: i'm suggesting changing the default values to give the user a better idea of what should go there.

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigDeployer.java
```diff
@@ -115,11 +115,21 @@ public void deployDefaultConfig(Path configDir, Path dataDir, Path pluginDir)
                 // TODO: Include node.id in install requests
                 updateProperty(configDir.resolve("node.properties"),
                         "node.id", randomUUID().toString());
+                /**
```
rschlussel: nit: all comments within a method should be in the same style.  so use `//`  here

presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoConfigDeployer.java
rschlussel: It shouldn't be the case that the user can specify something in a config file and it will always be ignored.  I think we shouldn't merge this commit and instead file a ticket for having a single source of truth for those properties (e.g. instead of specifying those properties in presto-manager, presto-manager reads them from the config files).

PR #63, opened 2017-08-16
Improve isRunning method

sudarsanm94:
Improve `isRunning()` method that checks whether Presto is running.

### DIFF
--- a/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInformer.java
+++ b/presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInformer.java
@@ -16,7 +16,6 @@ package com.teradata.prestomanager.agent;
 import com.google.inject.Inject;
 import io.airlift.log.Logger;
 
-import javax.ws.rs.ProcessingException;
 import javax.ws.rs.client.Client;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.UriBuilder;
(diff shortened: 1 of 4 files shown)


presto-manager-agent/src/main/java/com/teradata/prestomanager/agent/PrestoInformer.java
```diff
@@ -52,21 +51,17 @@ public int getPrestoPort()
 
     /**
      * Returns whether Presto is running as a coordinator
-     *
+     *<p>
+     * Always check whether Presto is running using {@link PackageController#isRunning()}
```
rschlussel: Why don't you call isRunning() from inside the method instead?

sudarsanm94: Because of two reasons:
1) `isRunning()` throws `PrestoManagerException`
I don't want `PrestoInformer` methods to throw `PrestoManagerException`
2) I don't want `isRunningCoordinator()` to return false if Presto itself is not running. I think it would be misleading.

rschlussel: You don't need to return false, but if a method is making assumptions, it's good to put in a `checkState()` that those assumptions hold.  You can catch the exception from isRunning() so that you don't need to rethrow it, and fail the checkState if you either encounter an exception or `isRunning()` returns false.

